Issue #2, opened 2019-05-08
constant folding


mwillsey:
We should implement constant folding. The `Language` trait should provide the functionality to do that.

remysucre:
Is pruning code working? I'm thinking to follow the `prune` function for folding. That is, match on nodes that are `op const const ...`, evaluate them, then add in the result as a node.

remysucre:
FYI I'm working on adding an `eval` method to the `Language` trait and then implement folding as above.

remysucre:
Hmm this isn't as simple as I thought, since the matcher seems necessary here. The suggestion by @ztatlock in #1 could help, but it's a bit weird to have `union` handle constants.

ztatlock:
Doing it in `union` is really just to incrementalize; it could be a separate pass too.  That said, there are a lot of bottom-up passes that can generically be incrementalized through `union` so worth considering?

PR #4, opened 2019-05-13
draft constant folding

remysucre:
draft code for constant folding #2

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -202,6 +202,54 @@ impl<L: Language> EGraph<L> {
         pruned
     }
 
+    fn get_children(ns: &Vec<Expr<L,Id>>) -> Vec<RecExpr<L>> {
+        let mut new_ns = Vec::new();
+        for n in ns {
+            if let Expr::Constant(c) = n {
+                let st : RecExpr<L> = RecExpr::from(Expr::Constant(c.clone()));
+                new_ns.push(st);
+            } else {
+                panic!("impossible")
+            }
+        }
+        new_ns
+    }
+
+    pub fn fold_constants(&mut self) -> usize {
+        let mut folded = 0;
+        for class in self.classes.values_mut() {
+            let mut new_nodes = Vec::new();
+            for node in &class.nodes {
+                match node {
+                    Expr::Operator(op, nids) =>
+                        if nids.iter().all(|x| true) {
+                            let ns = Vec::new();
+                            for nid in nids {
+                                // TODO
+                            }
+                            let ne = Expr::Operator(op.clone(), Self::get_children(ns));
+                            let nre = RecExpr::from(ne);
+                            let v = L::eval(&nre);
+                            new_nodes.push(Expr::Constant(v));
+                        } else {()},
+                    _ => (),
+                }
+            }
+
+            if new_nodes.len() > 0 {
+                folded += class.len() - new_nodes.len();
+                class.nodes = new_nodes;
+            }
+        }
+
+        if folded > 0 {
+            info!("Folded {} nodes", folded);
+        }
+
+        folded
+    }
+
+
     fn rebuild_once(&mut self) -> usize {
         let mut new_nodes = HashMap::default();
         let mut to_union = Vec::new();
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -1,4 +1,5 @@
 use log::*;
+use std::iter::ExactSizeIterator;
 use std::time::Instant;
 
 use crate::{
@@ -10,10 +11,9 @@ use crate::{
 /// Data structure to keep track of equalities between expressions
 #[derive(Debug)]
 pub struct EGraph<L: Language> {
-    // TODO no pub
-    pub nodes: HashMap<Expr<L, Id>, Id>,
-    pub leaders: UnionFind,
-    pub classes: HashMap<Id, EClass<L>>,
+    nodes: HashMap<Expr<L, Id>, Id>,
+    leaders: UnionFind,
+    classes: HashMap<Id, EClass<L>>,
     unions_since_rebuild: usize,
 }
 
@@ -51,7 +51,7 @@ pub struct AddResult {
 
 #[derive(Debug, Clone)]
 pub struct EClass<L: Language> {
-    id: Id,
+    pub id: Id,
     nodes: Vec<Expr<L, Id>>,
 }
 
@@ -59,6 +59,7 @@ impl<L: Language> EClass<L> {
     fn new(id: Id, nodes: Vec<Expr<L, Id>>) -> Self {
         EClass { id, nodes }
     }
+
     pub fn len(&self) -> usize {
         self.nodes.len()
     }
@@ -67,7 +68,7 @@ impl<L: Language> EClass<L> {
         self.len() == 0
     }
 
-    pub fn iter(&self) -> impl Iterator<Item = &Expr<L, Id>> {
+    pub fn iter(&self) -> impl ExactSizeIterator<Item = &Expr<L, Id>> {
         self.nodes.iter()
     }
 
@@ -97,6 +98,10 @@ impl<L: Language> EGraph<L> {
     pub fn add_expr(&mut self, expr: &RecExpr<L>) -> Id {
         let e = expr.as_ref().map_children(|child| self.add_expr(&child));
         self.add(e).id
+    }
+
+    pub fn classes(&self) -> impl ExactSizeIterator<Item = &EClass<L>> {
+        self.classes.values()
     }
 
     fn check(&self) {
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -15,6 +15,7 @@ pub struct EGraph<L: Language> {
     leaders: UnionFind,
     classes: HashMap<Id, EClass<L>>,
     unions_since_rebuild: usize,
+    debug: bool,
 }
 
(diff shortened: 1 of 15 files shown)


tests/prop.rs
```diff
@@ -40,6 +56,17 @@ impl Language for Prop {
...
         unimplemented!()
     }
+    fn eval(e: &RecExpr<Prop>) -> Bool {
+        match e.as_ref() {
+            Expr::Variable(_) => Bool::False, // TODO handle exception
+            Expr::Constant(c) => c.clone(),   // TODO should this be clone?
+            Expr::Operator(op, ns) => match op {
+                Op::And => ns.iter().map(Self::eval).fold(Bool::True, conjoin),
+                Op::Or => ns.iter().map(Self::eval).fold(Bool::False, disjoin),
+                _ => Bool::False, // TODO handle exception
```
mwillsey: It might be easier to assert that the operator has the right number of children, and then just unpack it and implement it that way. Ex:
```rust
Op::Not => {
  assert_eq!(ns.len(), 1);
  if ns[0] == Bool::True { Bool::False } else { Bool::True }
}

File: src/egraph.rs [hunk 2]
```diff
@@ -65,6 +65,20 @@ impl<L: Language> EClass<L> {
...
+
+    pub fn get_constant(&self) -> Option<RecExpr<L>> {
+        let mut res = None;
+        for n in &self.nodes {
+            match n {
+                Expr::Constant(c) => {
+                    res = Some(RecExpr::from(Expr::Constant(c.clone())));
+                    break;
+                }
+                _ => continue,
```
mwillsey: Excess control flow. Just `return` when when you find one.

src/egraph.rs
```diff
@@ -65,6 +65,20 @@ impl<L: Language> EClass<L> {
         self.id = id;
         self.nodes = more_nodes;
     }
+
+    pub fn get_constant(&self) -> Option<RecExpr<L>> {
```
mwillsey: If you change the type of `eval` (see below), then this should return `Option<L::Constant>`.

src/egraph.rs
```diff
@@ -202,6 +216,51 @@ impl<L: Language> EGraph<L> {
         pruned
     }
 
+    pub fn fold_constants(&mut self) -> usize {
+        let mut folded = 0;
+        let mut class_new_node = HashMap::default();
```
mwillsey: Perhaps call this `to_add`?

src/egraph.rs
```diff
@@ -202,6 +216,51 @@ impl<L: Language> EGraph<L> {
...
+                                break;
+                            }
+                        }
+                        // evaluate if all children are constant
+                        if all_consts {
+                            let ne = Expr::Operator(op.clone(), consts);
+                            let v = L::eval(&RecExpr::from(ne));
+                            let const_e = Expr::Constant(v);
+                            class_new_node.insert(*id, const_e);
+                            break;
```
mwillsey: What's the point of breaking here? Based on the invariant that you shouldn't have two constants in the same class?

I don't think we should do anything about that in this function. We can assert it later. Plus breaks are ugly :stuck_out_tongue:

src/egraph.rs
```diff
@@ -202,6 +216,51 @@ impl<L: Language> EGraph<L> {
...
+                    }
+                    _ => {}
+                }
+            }
+        }
+        // add and merge the new folded constants
+        for (cid, new_node) in class_new_node {
+            let aid = self.add(new_node).id;
+            self.union(cid, aid);
+            folded += 1;
```
mwillsey: declare folded down here, closer to use.

tests/prop.rs
```diff
@@ -147,3 +174,23 @@ fn prove_chain() {
...
+    let start_expr = Prop.parse_expr(start).unwrap();
+    assert_eq!(Prop::eval(&start_expr), Bool::False);
+}
+
+#[test]
+fn const_fold() {
+    let start = "(| (& F T) (& T F))";
+    let start_expr = Prop.parse_expr(start).unwrap();
+    let (mut eg, _) = EGraph::from_expr(&start_expr);
+    for n in 1..4 {
```
mwillsey: Why 4? This should probably assert that you eventually stop folding.

src/egraph.rs
mwillsey: Also, only add to folded if the result of the add wasn't there. Otherwise you can't tell if `fold_constants` hits fixpoint without pruning.

src/expr.rs
```diff
@@ -125,6 +125,8 @@ pub trait Language: Debug + PartialEq + Eq + Hash + Clone {
     type Wildcard: Debug + PartialEq + Eq + Hash + Clone;
 
     fn cost(node: &Expr<Self, u64>) -> u64;
+    // TODO change to return an Option, or just an Expr
+    fn eval(node: &RecExpr<Self>) -> Self::Constant;
```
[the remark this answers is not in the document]
remysucre: Ah I like that, I'll change it.

tests/prop.rs
remysucre: I just put the test there to look at the dot files. How do you assert folding stops? Just loop until fix point?

src/egraph.rs
remysucre: Asserting only one constant sounds good.

src/egraph.rs
remysucre: You mean instead of `class_new_node `?

tests/prop.rs
```diff
@@ -19,6 +19,22 @@ enum Bool {
...
+}
+
+fn disjoin(x: Bool, y: Bool) -> Bool {
+    if (x == Bool::True) || (y == Bool::True) {
+        Bool::True
+    } else {
+        Bool::False
+    }
+}
+
```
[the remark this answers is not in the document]
remysucre: Wouldn't that return a `bool` instead of the `Bool` you defined?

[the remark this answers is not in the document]
mwillsey: Whoops, you’re right.

tests/prop.rs
mwillsey: I’d unroll the loop and assert the return value of fold_constants (and maybe prune) until it’s 0.

Note that you’ll have to make one of the other fixes first, so that fold_constants actually returns 0.

remysucre: Got it

src/egraph.rs
```diff
@@ -202,6 +216,50 @@ impl<L: Language> EGraph<L> {
         pruned
     }
 
+    pub fn fold_constants(&mut self) -> usize {
+        let mut folded = 0;
+        let mut to_add = HashMap::default();
+        let mut constant_nodes = HashMap::default();
```
mwillsey: Filling up `constant_nodes` in the loop below might be too late. It introduces an ordering problem. I think you want two passes over the eclasses here. One to add the constants, and another to do the folding.

remysucre: How about the code now?

src/egraph.rs
remysucre: To be clear, should I add the assert somewhere or just do nothing for now?

src/egraph.rs
```diff
@@ -202,6 +202,58 @@ impl<L: Language> EGraph<L> {
...
+                        if let Some(c) = constant_nodes.get(id) {
+                            Some(c.clone())
+                        } else {
+                            let class = self.get_eclass(*id);
+                            for n in &class.nodes {
+                                if let Expr::Constant(c) = n {
+                                    return Some(c.clone());
+                                }
+                            }
+                            None
```
mwillsey: The else case here should never fire, right? If it was in the eclass it would have been added.

src/egraph.rs
```diff
@@ -202,6 +202,58 @@ impl<L: Language> EGraph<L> {
...
+        let mut folded = 0;
+        let mut to_add = HashMap::default();
+        // keep track of constants in each class to save some scanning
+        let mut constant_nodes = HashMap::default();
+
+        // look for constants in each class
+        for (id, class) in &self.classes {
+            for node in &class.nodes {
+                if let Expr::Constant(c) = node {
+                    constant_nodes.insert(*id, c.clone());
```
mwillsey: This still raises the question of what happens if there is more than one constant in an eclass.

Maybe for now the right thing to do is to just assert that there's only one.

src/egraph.rs
```diff
@@ -202,6 +202,58 @@ impl<L: Language> EGraph<L> {
...
+        }
+
+        // evaluate foldable expressions
+        for (id, class) in &self.classes {
+            for node in &class.nodes {
+                if let Expr::Operator(op, cids) = node {
+                    // get children if they are all constant
+                    let children: Option<Vec<_>> = cids.iter().map(|id| {
+                        if let Some(c) = constant_nodes.get(id) {
+                            Some(c.clone())
```
mwillsey: You can use [`Option::clone`](https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned) to clone the inside of an Option.

src/egraph.rs
mwillsey: See below, I think the right place to assert is while building the id-to-constant hashmap

src/egraph.rs
mwillsey: To be clear, I think this whole lambda can be replaced with:
```rust
|id| constant_nodes.get(id).cloned()
```

PR #11, opened 2019-09-25
egg-math: Add create_egraph and destroy_egraph C interfaces

oflatt:
This is the beginning of an interface that could be used by Herbie.
It's a rough PR but seems to be the right direction.


oflatt:
Looking into build failure now.

oflatt:
This code is out of date- it will be more effective to open a new pr later.

PR #12, opened 2019-10-01
Herbie egg interface

oflatt:
Here is a rough version of a C interface that Herbie (or other projects) could use.


oflatt:
egg-herbie will be its own fork, so this won't be merged.

PR #13, opened 2020-01-13
Add apply_random to randomly apply rewrite matches

remysucre:
`apply_random` samples a given number of matches to apply per rule per iteration. This can help avoid certain rules (e.g. associativity, commutativity) exploding after a few iterations, and also give each rule more equal opportunity to grow the graph.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ itertools = "0.8"
 smallvec = "0.6"
 indexmap = "1"
 instant = "0.1"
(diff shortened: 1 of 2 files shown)


mwillsey:
I think I'm going to close this for now in favor of the upcoming rewrite API improvements. They should give the user the ability to write something like this without messing with egg's code.

Although, this does seem generally useful. Perhaps we can provide a default implementation of it using the new API.

PR #21, opened 2020-04-16
Matching- skip eclasses that do not contain eclass with operator

oflatt:
This PR adds a hash table from the "signature" of the top level of a rule, which is the operator and the number of children, to eclasses that contain that signature at least once. For a given rule, it allows us to skip over eclasses that could not contain any matches for the rule because they do not contain an enode of the correct signature.

Timing with the rules from Herbie, this leads to a 3X speedup in searching (1.5X overall speedup) regardless of the node limit.
@mwillsey , could you please run whatever benchmarking you usually do on this PR and see what you find?

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -127,6 +127,7 @@ pub struct EGraph<L, M> {
     memo: IndexMap<ENode<L>, Id>,
     classes: UnionFind<Id, EClass<L, M>>,
     unions_since_rebuild: usize,
+    pub(crate) sigtoclasses: IndexMap<(L, usize), Vec<Id>>,
 }
 
 // manual debug impl to avoid L: Language bound on EGraph defn
@@ -147,6 +148,7 @@ impl<L, M> Default for EGraph<L, M> {
             memo: IndexMap::default(),
             classes: UnionFind::default(),
             unions_since_rebuild: 0,
+	    sigtoclasses: IndexMap::default(),
         }
     }
 }
@@ -327,6 +329,10 @@ impl<L: Language, M: Metadata<L>> EGraph<L, M> {
             #[cfg(feature = "parent-pointers")]
             parents: IndexSet::new(),
         };
+	self.sigtoclasses.entry((enode.op.clone(), enode.children.len()))
+	    .and_modify(|ids| ids.push(class.id))
+	    .or_insert(vec![class.id]);
+	
         M::modify(&mut class);
         let next_id = self.classes.make_set(class);
         trace!("Added  {:4}: {:?}", next_id, enode);
@@ -406,9 +412,21 @@ impl<L: Language, M: Metadata<L>> EGraph<L, M> {
     }
 
     fn rebuild_classes(&mut self) -> usize {
+	let (find, mut_values) = self.classes.split();
+
+	let mut sum: usize = 0;
+	for ids in self.sigtoclasses.values_mut() {
+	    sum += ids.len();
+	    let unique: IndexSet<Id> = ids
+		.iter()
+		.map(|id| find(*id))
+		.collect();
+	    ids.clear();
+	    ids.extend(unique);
+	}
+	
         let mut trimmed = 0;
 
-        let (find, mut_values) = self.classes.split();
         for class in mut_values {
             let old_len = class.len();
 
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -127,7 +127,7 @@ pub struct EGraph<L, M> {
     memo: IndexMap<ENode<L>, Id>,
     classes: UnionFind<Id, EClass<L, M>>,
     unions_since_rebuild: usize,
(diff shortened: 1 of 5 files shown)


src/egraph.rs
```diff
@@ -147,6 +148,7 @@ impl<L, M> Default for EGraph<L, M> {
             memo: IndexMap::default(),
             classes: UnionFind::default(),
             unions_since_rebuild: 0,
+	    sigtoclasses: IndexMap::default(),
```
mwillsey, line 151: Stuff is not aligned, please run `cargo fmt` or it will fail CI.

src/egraph.rs
```diff
@@ -127,6 +127,7 @@ pub struct EGraph<L, M> {
     memo: IndexMap<ENode<L>, Id>,
     classes: UnionFind<Id, EClass<L, M>>,
     unions_since_rebuild: usize,
+    pub(crate) sigtoclasses: IndexMap<(L, usize), Vec<Id>>,
```
mwillsey, line 130: Can we rename this to `classes_by_op`?

src/egraph.rs
```diff
@@ -406,9 +412,21 @@ impl<L: Language, M: Metadata<L>> EGraph<L, M> {
...
+	for ids in self.sigtoclasses.values_mut() {
+	    sum += ids.len();
+	    let unique: IndexSet<Id> = ids
+		.iter()
+		.map(|id| find(*id))
+		.collect();
+	    ids.clear();
+	    ids.extend(unique);
+	}
+
```
mwillsey, line 427: I think this is why it's failing tests. This doesn't capture enodes added during a metadata `modify`. This should probably just scan the eclasses (perhaps simultaneously with the loop below), rebuilding the map entirely.

src/pattern.rs
```diff
@@ -144,10 +144,22 @@ where
...
+		    vec![]
+		}
+	    },
+	    PatternAst::Var(v) => {
+		egraph
+		    .classes()
+		    .filter_map(|e| self.search_eclass(egraph, e.id))
+		    .collect()
+	    }
+	}
```
mwillsey, line 162: Could be put this logic in `machine` instead? It should make it even faster, because it will apply on recursive searches as well.

mwillsey: ah nvm, I think I see. It makes more sense here, since machine will only look at one eclass at a time.

src/egraph.rs
oflatt: Good idea!

src/egraph.rs
oflatt: Oh yes, I forgot to fix that. Thanks

src/pattern.rs
oflatt: Good idea, I'll try that

oflatt: I might be able to do that if I make the classes_by_op refer to a set

src/egraph.rs
oflatt: Rebuilding the map entirely would mean that you can only search directly after rebuilding. Right now rebuilding is optional before you search.
Would making an add_enode function for modify work better?

oflatt: (kind of optional, you don't get the congruence guarantee)

oflatt: Ah but then modify couldn't prune the nodes. I'll change it then.

src/pattern.rs
mwillsey: I worry that might be slower (at least for some eclasses), because the machine could either just look at the eclass instead

src/egraph.rs
mwillsey: rebuilding is not optional before you search, so you can assume it's been called

src/pattern.rs
oflatt: I think that's probably the case based on my other experiments

mwillsey:
My benchmarks:

master
```
benching diff_power_harder for 10 seconds...
bench    diff_power_harder:
  n = 11
  μ = 0.9918599499999999
  σ = 0.07366492475522622
Runner report
=============
  Stop reason: IterationLimit(50)
  Iterations: 50
  Rebuilds: 291633, 5832.66 per iter
  Total time: 0.9674673519999998
    Search:  (0.39) 0.374357273
    Apply:   (0.09) 0.09078622899999998
    Rebuild: (0.52) 0.502241622
```

this pr
```
benching diff_power_harder for 10 seconds...
bench    diff_power_harder:
  n = 13
  μ = 0.8230608773846154
  σ = 0.015520025769429393
Runner report
=============
  Stop reason: IterationLimit(50)
  Iterations: 50
  Rebuilds: 304051, 6081.02 per iter
  Total time: 0.817320584
    Search:  (0.25) 0.20449594200000004
    Apply:   (0.11) 0.08975624199999997
    Rebuild: (0.64) 0.523041465
```

PR #38, opened 2020-07-18
Efficient version of egraph.add_expr and extractor.find_best

yycdavid:
Add memoization to egraph.add_expr and extractor.find_best. On larger graphs, this gives a lot of speed up (changing from exponential to linear time). 

Run time comparison, on benchnet from https://github.com/yycdavid/tamago (~200 nodes):
-- add_expr: 
old version 1747s
new version 2.26s
-- find_best:
old version: 1177s
new version: 0.00123s


mwillsey:
Looks awesome, nice find! Also, thanks for fixing within the existing API. Excited to have this in the next release!

PR #62, opened 2021-03-17
Increase rank separation in graphviz output

fabric-and-ink:
At least in my use cases the graphviz output tends to get very cluttered quickly. I propose to increase the rank separation by default. This will likely not fit all uses cases, I am sure, but at least for me seems to be a good default value for non-trivial cases. Please let me know what you think.

### DIFF
--- a/src/dot.rs
+++ b/src/dot.rs
@@ -164,6 +164,8 @@ where
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         writeln!(f, "digraph egraph {{")?;
 
+        // Increase space between ranks to improve readability
+        writeln!(f, "  ranksep=2")?;
         // set compound=true to enable edges to clusters
         writeln!(f, "  compound=true")?;
         writeln!(f, "  clusterrank=local")?;


fabric-and-ink:
I think that it might even make sense to change the layout to left-to-right and remove the arrow anchors. But that is a bigger change.

mwillsey:
This sounds like a good idea. Do you have before and after pictures?

fabric-and-ink:
Yes, see below. The first is the default style, the second is the style produced by this pull request. The third is a left-to-right layout.

**default**
grafik

**with increased rank separation**
grafik

**left-to-right** (my personal favourite)
grafik

mwillsey:
I think this commit should address this in a flexble way: https://github.com/egraphs-good/egg/commit/c0ea058d4d4ca34b70a28e1cb4e6974c2f657c6b

Now you can just add whatever lines you want to the dotfile. Feel free to reopen if this is insufficient.

fabric-and-ink:
I think that this is a good solution. In addition, I propose to remove the anchors, since these will lead to suboptimal endpoint placement of the edges. Here is an example.

grafik

mwillsey:
I don't think we can remove the anchors in general, since the order of children is semantically important. Thoughts?

fabric-and-ink:
I solved a similar issue by putting a number label on the respective edge to indicate the position in the respective expression.

mwillsey:
The latest commit allows you to turn the anchors off.

fabric-and-ink:
Thank you! This will be very useful. Up until you said so, I was not aware that the argument position was retained in the graphs and I have been scratching my head about the anchors.

fabric-and-ink:
Very nice :heart: 

grafik

PR #68, opened 2021-03-27
Use new const generic rust feature to implement trait

fabric-and-ink:
This generalizes the trait implementation for all n. This would raise the minimum Rust compiler version to 1.51. Not sure if it is worth it. I stumbled over the impl macro and thought, hey that would fit!

### DIFF
--- a/src/language.rs
+++ b/src/language.rs
@@ -171,21 +171,27 @@ pub trait LanguageChildren {
     fn as_mut_slice(&mut self) -> &mut [Id];
 }
 
-macro_rules! impl_array {
-    () => {};
-    ($n:literal, $($rest:tt)*) => {
-        impl LanguageChildren for [Id; $n] {
-            fn len(&self) -> usize                   { <[Id]>::len(self) }
-            fn can_be_length(n: usize) -> bool       { n == $n }
-            fn from_vec(v: Vec<Id>) -> Self          { Self::try_from(v.as_slice()).unwrap() }
-            fn as_slice(&self) -> &[Id]              { self }
-            fn as_mut_slice(&mut self) -> &mut [Id]  { self }
-        }
-        impl_array!($($rest)*);
-    };
-}
+impl<const N: usize> LanguageChildren for [Id; N] {
+    fn len(&self) -> usize {
+        N
+    }
 
-impl_array!(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,);
+    fn can_be_length(n: usize) -> bool {
+        n == N
+    }
+
+    fn from_vec(v: Vec<Id>) -> Self {
+        Self::try_from(v.as_slice()).unwrap()
+    }
+
+    fn as_slice(&self) -> &[Id] {
+        self
+    }
+
+    fn as_mut_slice(&mut self) -> &mut [Id] {
+        self
+    }
+}
 
 #[rustfmt::skip]
 impl LanguageChildren for Box<[Id]> {


fabric-and-ink:
Not sure why the test fails... locally it fails even without my patch... will check tomorrow.

fabric-and-ink:
Ok, strange. When I run `cargo test` locally, the test `test lambda_fib::test` fails. Even for the main branch, older commits and older rust versions. The test pipeline for this pull request fails however since it is not using rust 1.51.

Dandandan:
@fabric-and-ink are you testing with release?
`cargo test --release` or `make test`?
some tests time out when it takes

fabric-and-ink:
I ran `cargo test`. You are right, there seems to be a timeout. To be honest, I think it is not a good idea to put a timeout in those tests because then whether or not they succeed depends on my machine and compiler optimizations. Here is the log.

```
---- lambda_fib::test stdout ----
Runner report
=============
  Stop reason: TimeLimit(6.1912622)
  Iterations: 22
  Egraph size: 3577 nodes, 1318 classes, 8278 memo
  Rebuilds: 1493, 67.86 per iter
  Total time: 6.212390999999999
    Search:  (0.91) 5.6573369
    Apply:   (0.05) 0.28797360000000005
    Rebuild: (0.04) 0.2670052
End (33): (app
  (fix
    fib
    (lam
      n
      (if
        (= (var n) 0)
        0
        (if
          (= (var n) 1)
          1
          (+ (app (var fib) (+ (var n) -1)) (app (var fib) (+ (var n) -2)))))))
  4)
Trying to prove goal 0: 3
thread 'lambda_fib::test' panicked at 'Could not prove goal 0:
3
Best thing found:
(app
  (fix
    fib
    (lam
      n
      (if
        (= (var n) 0)
        0
        (if
          (= (var n) 1)
          1
          (+
            (app (var fib) (+ (var n) -1))
            (app (var fib) (+ (var n) -2)))))))
  4)', C:\git\egg\src\egraph.rs:331:17

---- lambda_function_repeat::test stdout ----
Runner report
=============
  Stop reason: TimeLimit(20.173931)
  Iterations: 42
  Egraph size: 35563 nodes, 15555 classes, 63151 memo
  Rebuilds: 8551, 203.60 per iter
  Total time: 20.361541000000003
    Search:  (0.78) 15.848708300000002
    Apply:   (0.09) 1.8312556000000002
    Rebuild: (0.13) 2.6812234
End (12): (lam x (+ 1 (let x (var x) (+ (var x) 1))))
Trying to prove goal 0: (lam ?x (+ (var ?x) 2))
thread 'lambda_function_repeat::test' panicked at 'Could not prove goal 0:
(lam ?x (+ (var ?x) 2))
Best thing found:
(lam
  x
  (+ 1 (let x (var x) (+ (var x) 1))))', C:\git\egg\src\egraph.rs:331:17


failures:
    lambda_fib::test
    lambda_function_repeat::test
```

src/language.rs
```diff
@@ -171,22 +171,24 @@ pub trait LanguageChildren {
...
-            fn can_be_length(n: usize) -> bool       { n == $n }
-            fn from_vec(v: Vec<Id>) -> Self          { Self::try_from(v.as_slice()).unwrap() }
-            fn as_slice(&self) -> &[Id]              { self }
-            fn as_mut_slice(&mut self) -> &mut [Id]  { self }
-        }
-        impl_array!($($rest)*);
-    };
+impl<const N: usize> LanguageChildren for [Id; N] {
+    fn len(&self) -> usize {
+        <[Id]>::len(self)
```
mwillsey, line 176: This could just be `N`, right?

mwillsey:
I'm fine with relying on latest rust. Let me bump CI first. I'll also disable the longer running tests when not in release mode.

src/language.rs
fabric-and-ink: Right! I missed that.

fabric-and-ink:
Is the limit to 16 documented anywhere? I would remove that note later.

mwillsey:
I don't think it is.

mwillsey:
Ok CI should be good to go for 1.51. Can you rebase and fix the N thing? Otherwise LGTM!

fabric-and-ink:
Pipeline is happy now.

PR #71, opened 2021-04-08
Print eclass data in egraph dump

fabric-and-ink:
Print eclass data in egraph dump to help debugging.


mwillsey:
Good idea! Should you maybe print the data before the colon? At the end seems like it might be noisy when there are a lot of nodes.

fabric-and-ink:
I moved it before the colon. Not sure how big data fields typically get... Let me know what you think :)

PR #74, opened 2021-04-11
Improved e-matching for ground terms

yihozhang:
According to the existing query compilation, every enode will be compiled to a `Bind` operator, which will always add one layer of non-determinism. However, it turns out this is unnecessary.

Consider the example `(+ ?x (+ a b))`, current compilation algorithm will make the program to search for all `+`-app enode `e`, and for each enode in `e.children[1]`, search for the pattern `(+ a b)`. However, because the eclass `(+ a b)` always maps to the same eclass, we can have a simple `Compare` operator to check for eclass equality.

This PR introduces the `Load` operator, which will pop the last `node.len()`-many  Id in the register and push to the register the corresponding enode. Under this compilation algorithm, all ground patterns will be loaded once before the actual e-matching and will be used for comparison. Finally, during compilation, ground patterns are given higher priority and will thus be considered first.

It reduces the run time of the following query from 11.2s (resp. 10.4s) to 9.95ms (resp. 1.12s).
```rust

#[test]
fn math_associate_adds_harder() {
    let rules = &rules();
    let runner: Runner<Math, _> = Runner::default()
        .with_iter_limit(12)
        .with_node_limit(100_000)
        .with_time_limit(std::time::Duration::from_secs(100))
        .with_scheduler(SimpleScheduler)
        .with_expr(
            &"(+ (+ (+ (+ a1 a2) (+ a3 a4)) (+ (+ a5 a6) (+ a7 a8))) (+ (+ a9 a10) (+ a11 a12)))"
            // &"(+ (+ (+ (+ (+ (+ (+ (+ (+ (+ a1 a2) a3) a4) a5) a6) a7) a8) a9) a10) a11)"
                .parse()
                .unwrap(),
        )
        .run(rules);
    runner.print_report();
    let pattern: Pattern<Math> =
        "(+ (+ (+ (+ a1 a2) (+ a3 a4)) (+ (+ a5 a6) (+ a7 a8))) (+ (+ a9 a10) (+ a11 a12)))"
            // "(+ (+ (+ (+ (+ (+ (+ (+ (+ (+ a1 a2) a3) a4) a5) a6) a7) ?z1) ?z) ?y) ?x)"
            .parse()
            .unwrap();
    let start_time = std::time::Instant::now();
    let result = pattern.search(&runner.egraph);
    println!("time elapsed {:?}", std::time::Instant::now() - start_time);
    println!(
        "match size {}",
        result.iter().map(|m| m.substs.len()).sum::<usize>()
    );
}
```

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -257,6 +257,18 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
         id.map(|&id| self.find(id))
     }
 
+    /// Lookup the eclass of the given [`RecExpr`].
+    pub fn lookup_expr(&self, expr: &RecExpr<L>) -> Option<Id> {
+        let nodes = expr.as_ref();
+        let mut new_ids = Vec::with_capacity(nodes.len());
+        for node in nodes {
+            let node = node.clone().map_children(|i| new_ids[usize::from(i)]);
+            let id = self.lookup(node)?;
+            new_ids.push(id)
+        }
+        Some(*new_ids.last().unwrap())
+    }
+
     /// Adds an enode to the [`EGraph`].
     ///
     /// When adding an enode, to the egraph, [`add`] it performs
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -26,7 +26,6 @@ enum Instruction<L> {
     Bind { node: L, i: Reg, out: Reg },
     Compare { i: Reg, j: Reg },
     Load { node: L, out: Reg },
-    // Check { node: L, i: Reg },
 }
 
 #[inline(always)]
@@ -208,11 +207,11 @@ impl<'a, L: Language> Compiler<'a, L> {
                     }
                 }
                 ENodeOrVar::ENode(node) => {
-                    // if node.all(|c| is_const[usize::from(c)]) {
-                    //     self.load_term(node, &mut instructions, self.out);
-                    //     instructions.push(Instruction::Compare { i, j: self.out });
-                    //     continue;
-                    // }
+                    if node.all(|c| is_const[usize::from(c)]) {
+                        self.load_term(node, &mut instructions, self.out);
+                        instructions.push(Instruction::Compare { i, j: self.out });
+                        continue;
+                    }
 
                     let out = self.out;
                     self.out.0 += node.len() as u32;
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -1,5 +1,5 @@
 use crate::util::HashSet;
-use crate::{Analysis, EClass, EGraph, ENodeOrVar, Id, Language, PatternAst, Subst, Var};
+use crate::{Analysis, EClass, EGraph, ENodeOrVar, Id, Language, PatternAst, RecExpr, Subst, Var};
 use std::cmp::Ordering;
 
 struct Machine {
@@ -18,6 +18,7 @@ struct Reg(u32);
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Program<L> {
     instructions: Vec<Instruction<L>>,
+    ground_terms: Vec<RecExpr<L>>,
     subst: Subst,
 }
 
@@ -25,7 +26,6 @@ pub struct Program<L> {
 enum Instruction<L> {
     Bind { node: L, i: Reg, out: Reg },
     Compare { i: Reg, j: Reg },
-    Load { node: L, out: Reg },
 }
 
 #[inline(always)]
@@ -102,18 +102,6 @@ impl Machine {
                         return;
                     }
                 }
-                Instruction::Load { node, out } => {
-                    let mut iter = self.reg[out.0 as usize..].iter();
-                    let mut node = node.clone();
-                    node.update_children(|_| *iter.next().unwrap());
-                    match egraph.lookup(&mut node) {
-                        Some(id) => {
-                            self.reg.truncate(out.0 as usize);
-                            self.reg.push(id);
-                        }
-                        None => return,
-                    }
-                }
             }
         }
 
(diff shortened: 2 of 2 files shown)


mwillsey:
I love this idea. Can I maybe propose a simplifying suggestion? I don't think we actually need a `Load` instruction.

What if a `Program<L>` also contained `ground_terms: Map<RecExpr<L>, Reg>` or something, or perhaps even just `Vec<RecExpr<L>>`. And then execution of a machine first involves placing the ground terms in the appropriate registers on initalization (just bailing if you can't find them in the e-graph at all). This way there is no `Load` instruction, it basically just becomes a compare.

So your example `(+ ?x (+ a b))` would become 
```
ground_terms: vec![ (+ a b) ],
instructions: Bind(+, 1, 1), Compare(0, 2) [[ i think ]]
```

My registers might be off there. But the important change is how registers are initialized in the machine. Right now, the only thing that happens is reg0 becomes the e-class you are searching. Now the registers would be initialized to [ground_1, ..., ground_n, starting_eclass]. This shoudl also simplify the code because you can use `EGraph::lookup`.

yihozhang:
This looks good to me! One thing I may be missing is what do you mean by using `EGraph::lookup`? `lookup` seems to take to lookup for a enode instead of a `RecExpr` and I was unable to find a function in the interface that search for a `RecExpr`. Maybe we should implement and expose one as it may be useful in some other cases?

mwillsey:
Ah that's a good point! We should have a lookup_rec.

0x0f0f0f:
That's a clever optimization. Thanks!

philzook58:
This is a really interesting observation. That pattern matching can be considered from both a top down and bottom up perspective.

Pushing back on not needing an instruction if you generalize this slightly, suppose you had a pattern `f(?a , big_tree(?a))`. One could use top down pattern matching to find the first `?a` but then switch to bottom up mode to find the `big_tree(?a)` much more efficiently via lookup in the egraph. In general the goal might be the find the shallowest instances of variables interleaved with switching to bottom up mode. This may not be a common use of patterns though. I don't know

A speculative thought: It feels kind of possible to maintain partial pattern matches between ematching runs, since the monotonic nature of the egraph means they don't go away. This amounts to some kind of memoization of the pattern matching. In other words, the bottom up perspective could be taken as the primary one. In my exercises in writing ematching, it amounts to a very brute force search over the entire egraph every time. I don't know if egg has something smarter than that, It might even be possible to detect new partial pattern creation at time the of expression addition to the egraph or during rebuilding.

mwillsey:
@philzook58 we've got some big plans in the pipeline on how to do e-matching much more efficiently in such a way that hopefully renders most of this obsolete :). Long story short, we've come up with a relational encoding of e-graphs and patterns, and that will allow a lot more flexibility in the "order" in which you search (top-down, bottom up, middle out, whatever). You can a preliminary paper by @yihozhang here: http://effect.systems/doc/relational.pdf

For your second thought, there are two relevant threads. First, the original equality saturation paper used a Rete network to do e-matching, which does have this "partial match" thing you describe. Second, [the paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.800.2292&rep=rep1&type=pdf#page=193) that egg's current e-matching is based on talks about the mod-time optimization, which is kinda relevant.

egg uses neither of those because it's already very fast, those optimization are complex, and it's not even clear they are much faster. Plus, with the new way that we'd like to take e-matching, those will be irrelevant.

yihozhang:
I made some changes per @mwillsey 's suggestions two days ago and just made a pass through it. Let me know if there are any questions!. Meanwhile, I also think this PR will become obsolete in the future, but it would still be great to incorporate this and have a release before the brand new e-matching algorithm, any ideas?

0x0f0f0f:
I had the recurring thought of viewing e-graphs as matrices for a while, but this idea of using CQs and viewing f-apps as relational tables is very nice, and I haven't thought about it. One can also very efficiently pack these tables in matrices of small unsigned integers. Are you planning to see if this table representation can also speed up the add-merge-repair procedures?

mwillsey:
@0x0f0f0f I don't think this will have any effect on add-merge-rebuild. That part is not the bottleneck.

0x0f0f0f:
Just implemented this tiny improvement for Metatheory.jl. Thanks @yihozhang @mwillsey
On this Knuth-Bendix like test-case (groups):

```
function rep(x, op, n::Int)
    foldl((x, y) -> :(($op)($x, $y)), repeat([x], n))
end

Mid = @theory begin 
    a * :ε => a
    :ε * a => a
end 

Massoc = @theory begin
    a * (b * c) => (a * b) * c
    (a * b) * c => a * (b * c) 
end 

T = [
    @rule :b*:B => :ε
    @rule :a*:a => :ε
    @rule :b*:b*:b => :ε
    @rule :B * :B => :B
    RewriteRule(Pattern(rep(:(:a*:b), :*, 7)), Pattern(:(:ε)))
    RewriteRule(Pattern(rep(:(:a*:b*:a*:B), :*, 5)), Pattern(:(:ε)))
]

G = Mid∪Massoc∪T
expr = :(a*b* a*a*a * b*b*b * a * B*B*B*B * a)
```

It used to take 135 seconds without this ground term matching optimization.
It now takes...
0.5
:rocket: 

CC @philzook58

0x0f0f0f:
@philzook58 
> A speculative thought: It feels kind of possible to maintain partial pattern matches between ematching runs, since the monotonic nature of the egraph means they don't go away. This amounts to some kind of memoization of the pattern matching. 

This KB groups example now takes 0.120737137 seconds for 8 iterations of eqsat, by memoizing the eclasses corresponding to matched ground term patterns in the egraph.

PR #79, opened 2021-04-14
Add Send and Sync bounds to Rewrite so it can be shared between threads

rosekunkel:
The `searcher` and `applier` fields of `Rewrite` use `Arc`, but without `Send` and `Sync` bounds on the trait objects, they still can't be shared between threads. This also makes them incompatible with `lazy_static`, e.g. this code won't compile
```rust
lazy_static! {
    static ref REWRITE: Rewrite<MyLang, ()> = rewrite!("foo"; "bar" => "baz");
}
```


mwillsey:
Nice! Good catch and thanks for the PR :)

PR #81, opened 2021-04-16
Clarify Analysis Data lattice operations

bjchambers:
This introduces two basic lattice operations -- ordering and merging
of `Data` and uses those to describe (and implement) a combined "merge
and return ordering" operation.

The combined operation is what is actually used, allowing for additional
optimizations when it can be simultaneously implemented.

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -379,7 +379,12 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
         assert_eq!(id1, class1.id);
 
         self.pending.extend(class2.parents.iter().cloned());
-        match self.analysis.merge(&mut class1.data, class2.data) {
+
+        let class1_data = std::mem::take(&mut class1.data);
+        let (ord, data) = N::cmp_merge_data(class1_data, class2.data);
+        class1.data = data;
+
+        match ord {
             Some(Ordering::Equal) => {}
             Some(Ordering::Greater) => {
                 self.analysis_pending.extend(class2.parents.iter().cloned());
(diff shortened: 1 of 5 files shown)


bjchambers:
Passes tests (but not doctests). Benchmarks seem to be affected, but haven't re-run recently.

src/egraph.rs
```diff
@@ -516,12 +521,17 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
             while let Some((node, class_id)) = self.analysis_pending.pop() {
                 let class_id = self.find_mut(class_id);
                 let node_data = N::make(self, &node);
+
                 let class = self.classes.get_mut(&class_id).unwrap();
-                match self.analysis.merge(&mut class.data, node_data) {
-                    Some(Ordering::Equal) | Some(Ordering::Greater) => {}
+                let data = std::mem::take(&mut class.data);
```
bjchambers, line 526: I think the `take` is necessary, but it would be good if it wasn't. Specifically, while I'd like to be able to do this using the `entry` API on `self.classes`, I think that leads to problems -- we need to borrow the graph for the call to `N::modify(..)` as well as extending the `analysis_pending`, which needs to happen within the scope that the entry has already borrowed those.

mwillsey: I was able to solve this in my commit.

mwillsey:
So I was able to remove the `Data: Default` requirement, but I'm still left unsure whether or not this an improvement to the existing API. In the `lambda` example, it's still easier to use the "old" way, and implementing the `merge` or `compare` functions is neither simple nor necessary. In the other easier domains, its not that much easier than the old way.

Thoughts?

bjchambers:
> So I was able to remove the `Data: Default` requirement, but I'm still left unsure whether or not this an improvement to the existing API. In the `lambda` example, it's still easier to use the "old" way, and implementing the `merge` or `compare` functions is neither simple nor necessary. In the other easier domains, its not that much easier than the old way.
> 
> Thoughts?

I had some similar concerns. I think the largest benefits are:

1. Being able to describe the expected behavior in terms of more primitive lattice operations. Eg., `cmp` is lattice order, `a <= merge(a, b)` and `b <= merge(a, b)`, etc. This could probably be accomplished without separate methods, just by updating documentation. On the other hand, having separate methods would allow the use of `debug_assert!` to enforce some of those invariants, which may help find errors in the analysis.
2. Allowing the implementation of a few (simpler) methods to define the behavior. This doesn't seem to be a huge benefit in the current shape. Another option to provide them would be a method like `merge_min` that took the comparison and merge operator, or something like that.
3. (Possibly) if it made it easier to combine analysis parts. For instance, you may have two simple lattices (type and constant folding) and wish to combine them. This combination should be mostly routine.

mwillsey:
1. It seems like we can't actually offer this, because `merge` takes owned `Data`s and `Data` doesn't implement `Clone`. You can't compare merged result to the inputs because they are consumed.
2. Yeah, it doesn't seem to buy much right now. We could have `merge` just panic by default, which makes sense when your order is total. But implementing `cmp` isn't that much easier than the old `merge` in that case.
3. This is a great idea that I would love to implement at some point, but I think the old `merge` does work okay here. You can merge (A, B) by just merging into A and then B, and combining the ordering results that come out.

Maybe we should table this for now, and revisit when implementing composition of Analyses.

mwillsey:
Closing for now. Feel free to reopen!

PR #86, opened 2021-04-30
Move printing to `Display` and parsing to a new trait, `FromOp`.

### LINKED ISSUE #83: Language::display_op should work like other formatting traits
The rust std library formatting traits (`Display`, `Debug`, `Binary`, etc.) are all structured the same way: they have a single method
```rust
fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result
```
`Language::display_op` should follow the same convention, either by changing its signature:
```rust
fn display_op(&self, f: &mut fmt::Formatter) -> fmt::Result
```
or by separating its functionality out into a new trait:
```rust
pub trait DisplayOp {
    pub fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result;
}
```
Beyond just consistency, this change would also have the added advantage that the operator's string representation could be generated on-the-fly, whereas the current signature (returning `&dyn Display`) requires the implementer to return a reference to some value that already exists. I'm happy to submit a pull request making this change.

rosekunkel:
This fixes #83 by replacing `Language::display_op` with ~`OpStr::fmt`~ `Display`. It also moves `Language::from_op_str` to ~`OpStr::from_str`~ `FromOp::from_op`. Moving these two functions to a separate trait eliminates the need to give them default implementations which panic. `define_language!` has been updated to automatically implement ~`OpStr`~ `Display` and `FromOp`, so the transition should be seamless for people using the macro.

Additionally, this pull request adds a dependency on [thiserror](https://docs.rs/thiserror/1.0.24/thiserror/), a crate which provides a derive macro for `std::error::Error`, as this refactoring replaces several instances of `String` being used as an error type with structured error types implementing `Error`.

Let me know if there's anything that should be adjusted!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@
   [Downey, Sethi, Tarjan](https://dl.acm.org/doi/pdf/10.1145/322217.322228),
   and the analysis data propagation is more precise with respect to merging.
   Overall, the algorithm is simpler, easier to reason about, and more than twice as fast!
+- ([#86](https://github.com/egraphs-good/egg/pull/86))
+  `Language::display_op` and `Language::from_op_str` have been replaced by a new
+  `OpStr` trait. Their rough equivalents are `OpStr::fmt` and `OpStr::from_str`.
+  `define_language!` has been updated to generate `OpStr` implementations, so
+  this change should be mostly transparent to users of the macro.
 
 ## [0.6.0] - 2020-07-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,11 @@
   and the analysis data propagation is more precise with respect to merging.
   Overall, the algorithm is simpler, easier to reason about, and more than twice as fast!
 - ([#86](https://github.com/egraphs-good/egg/pull/86))
-  `Language::display_op` and `Language::from_op_str` have been replaced by a new
-  `OpStr` trait. Their rough equivalents are `OpStr::fmt` and `OpStr::from_str`.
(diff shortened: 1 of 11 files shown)


mwillsey:
Awesome! There's a lot of good stuff here. A couple thoughts:

- Why not just have Language implement Display? We could still have a separate parsing trait.
- I'm not ready for a thiserror dependency just yet, especially since we just have one error.

mwillsey:
Ah, I see your thiserror changes are actually pretty widespread. I guess it's worth keeping the. `thiserror` is the future after all

rosekunkel:
Good questions! Re: `Display`, do you mean that we should mandate that if a `Language` implements `Display`, it should just display the name of the operator?

As I've been thinking about this more, I'm wondering whether we really want to be using strings at all. Right now the way these string functions are used is for constructing and parsing sexps, so maybe it would make more sense if we had `ToSexp` and `FromSexp` traits instead.

Also, re: `thiserror`, it's just a convenience macro, so if we want to remove the dependency later we can just implement `Error` manually instead.

mwillsey:
Yes, Display just prints the operator.

That's true, it's mostly used to print sexps, but there are couple places where it's nice to just print the operators, like in Dot.

rosekunkel:
Ok, I think that makes sense. I'll rework it into a trait for parsing and `Display` for printing the operator.

rosekunkel:
@mwillsey I made the changes you suggested -- let me know if there's anything else I should change.

mwillsey:
Nice work, thanks!

PR #108, opened 2021-07-24
Change signature to simplify merge

bjchambers:


mwillsey:
Nice! I do like that they are composable. But I'm not totally sure that `bool` in the return type is much clearer. I share your concern that `merge` is currently a little awkward to write. But the returned `bool`s are a little subtle to understand.

In the PR you use the names `changed1` and `changed2`, which seems intuitive, but `b` is immutable! So it's not quite `changed` but `is_different_from_the_result`, or something. Plus, there is always the issue that `bool`s can be flipped, so the `merge` implementations write `!=` as opposed to `==`.

I would maybe advocate for flipping the bool. That way, instead of `changed`, its `is_merged` or something. So returning `(true, true)` means both are actually the merged value; they are the same.

Thoughts?

bjchambers:
That could work. I think either polarity works. What drew me to that was
when thinking about the composed case it is easier into think in terms of
"what is changed or unchanged" instead of the actual ordering. So flipping
the sign and updating the name preserves that benefit, since you just
need.to reason about how the input differs from the output. I could also
see renaming to `is_different` instead of changed. Happy to make such a
change.

We could also define an enum instead of bools. Something like `IsMerged`
and `IsDifferent`.

mwillsey:
Upon reflection, I actually like the code as is. 

Two thoughts before merging:

1. Can you enhance the `merge` docs to a give a case-wise spec as it did before?
2. Could we change the return type of `merge` to a `pub struct DidChange(pub bool, pub bool)`? That way the meaning is clear, and you could even implement `BitOr` on it to make composition clean!

bjchambers:
> Upon reflection, I actually like the code as is.
> 
> Two thoughts before merging:
> 
>     1. Can you enhance the `merge` docs to a give a case-wise spec as it did before?
> 
>     2. Could we change the return type of `merge` to a `pub struct DidChange(pub bool, pub bool)`, or maybe `DidMerge`? That way the meaning is clear, and you could even implement `BitOr` on it to make composition clean!

I like that -- especially the implementation of `BitOr`.

I considered adding constants like the following for documentation, but it seemed like it would obscure the fact that the result is made up of two separate, meaningful parts.

```
/// Constant indicating neither A nor B did change as part of a merge.
///
/// This may be the case if the analysis data represents a lattice and
/// `a` was already greater than `b`.
const UNCHANGED: Self = Self::new(false, false);
```

mwillsey:
I don't think we need constants, or even functions on the DidMerge type. Just the struct constructor `DidMerge(true, false)` should be fine.

Also, be careful! The value `(false, false)` indicates that no merging needs to be done! So, `a == b`, not `a >= b` as your comment claims.

bjchambers:
Added the struct / updated comments. I can switch it to a tuple-struct, but it currently uses fields, which seemed to provide a better path for documenting each part.

mwillsey:
I would prefer a tuple struct (with public fields) named `DidMerge`. I think that best conveys the meaning. Putting all the documentation on the struct itself (or better yet, on the `merge` method) seems good to me.

bjchambers:
Updated to use a tuple struct, and took a stab at the documentation. Let me know how that looks.

mwillsey:
Thanks for the PR!

PR #115, opened 2021-08-07
Egg Proof Production

oflatt:
This PR adds the ability to generate explanations for two terms being equivalent in the graph to egg. The `explain_equivalence` function on the runner provides an `Explanation` for two `RecExpr` being equal. This explanation has a tree format as well as a flattened format. In the tree format, children can be proofs. In the flattened format, there is only one rewrite happening between terms.

Users who which to perform unions themselves (either in custom appliers or in analyses) are required to use the `union_with_justification` function, which requires two patterns (lhs and rhs) and their substitution justifying the union. The `Explain` data structure internally finds enodes corresponding to the e-match for those patterns. 

Implementing proof-production involved some interesting design decisions. The biggest question is how to efficiently find the enodes corresponding to the user-given pattern. When the egraph congruence invariant is maintained, this is easy because the hash-cons allows us to find each node matching the pattern from bottom-up. So the solution in this PR is that whenever `union_with_justification` is called, the congruence invariant needs to be true.

Here's how that works:
- Instead of applying each union directly, unions are put on a `to_union` stack and delayed until rebuilding. So until rebuilding, the memo remains updated and calls to `union_with_justification` are valid.
- At rebuilding, we apply all the unions.
- Previously, maintenance of analysis was interweaved with rebuilding, but this won't work anymore because congruence has not been restored.
- So after rebuilding, we apply analysis results and rebuild again until fixpoint. But further analysis unions must also be delayed, so there is another stack for that. So far it's unclear what the performance hit of doing this is.

This PR isn't ready yet- I've posted it so @mwillsey and others can see it.
TODOS:
- [x] Documentation of functions
- [x] Documentation of proof generation, including explanations of explanation trees and flattened explanations
- [x] Make proof-generation flag work, including disallowing calling `union` directly in explanation-generation mode
- [x] Caching in proof-generation
- [x] Decided against adding Herbie benchmarks
- [x] Let bindings so you can reuse proofs in the printed output
- [x] Benchmarking how the change to rebuilding affects performance
- [x] Reduce overhead of proof production when feature is off- don't delay unions?
- [x] Change rewrite rule string to an Arc
- [x] Simplify union_with_justification api
- [x] Pass patternast as a Cow from apply_one
- [x] Optional- add tutorial


src/egraph.rs
```diff
@@ -46,10 +46,10 @@ and
 pub struct EGraph<L: Language, N: Analysis<L>> {
     /// The `Analysis` given when creating this `EGraph`.
     pub analysis: N,
+    pub explain: Explain<L>,
+    to_union: Vec<(Id, Id, Option<String>)>,
```
oflatt, line 50: This is the new to_union stack. When proof production is on, it is important that these id's are not canonical- they represent the enode-ids of the two enodes that are the match and application.

src/egraph.rs
```diff
@@ -510,19 +547,35 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
...
+
         let mut n_unions = 0;
 
         while !self.pending.is_empty() {
-            while let Some((mut node, class)) = self.pending.pop() {
-                node.update_children(|id| self.find_mut(id));
-                if let Some(memo_class) = self.memo.insert(node, class) {
-                    let (_, did_something) = self.union(memo_class, class);
-                    n_unions += did_something as usize;
+            while !self.pending.is_empty() {
```
oflatt, line 567: Here is the big change to rebuilding- we have to restore congruence completely before doing analysis stuff

src/rewrite.rs
```diff
@@ -147,6 +148,23 @@ where
             .collect()
     }
 
+    /// For patterns, return the ast directly as a reference
+    fn get_pattern_ast(&self) -> Option<&PatternAst<L>> {
```
oflatt, line 152: get_pattern_ast is here for simple patterns that don't depend on the context

src/rewrite.rs
```diff
@@ -147,6 +148,23 @@ where
...
+        None
+    }
+
+    /// Return the pattern corresponding to the searcher.
+    ///
+    /// This pattern is commonly used for proof-production mode.
+    fn get_ast(
```
oflatt, line 159: I'm a little bit unhappy with the API right now- often get_ast and apply_one have similar implementations. Not sure how to combine them nicely (see lambda.rs for an example)

src/rewrite.rs
```diff
@@ -256,46 +281,102 @@ where
...
+    ) -> Option<PatternAst<L>> {
+        None
+    }
+
+    /// Unions the eclasses after a match has been applied.
+    ///
+    /// This should return a list of [`Id`]s of eclasses which
+    /// had unions applied to them.
+    /// [`union_results`]: Applier::union_results()
+    fn union_results(
```
oflatt, line 341: `apply_matches` by default now calls union_results instead of unioning directly, giving users a hook to do the right thing in proof production mode

src/explain.rs
```diff
@@ -0,0 +1,806 @@
...
+            nodes.push(&self.explainfind[usize::from(node)]);
+            if next == ancestor {
+                return nodes;
+            }
+            assert!(next != node);
+            node = next;
+        }
+    }
+
+    fn explain_enodes(&self, left: Id, right: Id) -> ExplanationTrees<L> {
```
oflatt, line 745: This function is the whole proof-production algorithm. The expression corresponding to a node is guaranteed to match any rules it is connected to, so rewrites are easy to prove.

src/explain.rs
```diff
@@ -0,0 +1,806 @@
...
+        &mut self,
+        node1: Id,
+        node2: Id,
+        canon_id1: Id,
+        canon_id2: Id,
+        justification_maybe: Option<Justification>,
+    ) -> Id {
+        assert_ne!(self.find(node1), self.find(node2));
+        if let Some(justification) = justification_maybe {
+            self.make_leader(node1);
```
oflatt, line 678: Reverse edges in the explainfind so that we can add an edge from node1->node2

src/macros.rs
```diff
@@ -261,7 +261,7 @@ rules.extend(vec![
 #[derive(Debug)]
 struct MySillyApplier(&'static str);
 impl Applier<SimpleLanguage, ()> for MySillyApplier {
-    fn apply_one(&self, _: &mut EGraph, _: Id, _: &Subst) -> Vec<Id> {
+    fn apply_one(&self, _: &mut EGraph, _: Id, _: &Subst) -> (Vec<Id>, Option<PatternAst<SimpleLanguage>>) {
```
oflatt, line 264: Hmm, what if different applications have different PatternAst? Same for searchers.

tests/lambda.rs
```diff
@@ -39,15 +39,32 @@ struct LambdaAnalysis;
...
+        )),
+        Lambda::Bool(b) => 
+            Some((
+            enode.clone(),
+            format!("{}", b).parse().unwrap(),)),
+            
+        Lambda::Add([a, b]) =>
+            Some((
+            Lambda::Num(x(a)?.num()? + x(b)?.num()?),
+            format!("(+ {} {})", pat(a)?, pat(b)?).parse().unwrap(),
```
oflatt, line 61: There's a very subtle problem here- we would like to insert simply pattern corresponding to (+ constant constant)
But the class analysis may not have unioned the constant with the class yet due to delayed unions
So we have to recursively build the pattern from the children, which is bad. Not sure how to fix this yet

oflatt:
Update: Rebuilding with analysis is still broken in some way (lambda_function_repeat runs out of memory) but I think I can get it right

oflatt:
I added caching, and the lambda_function_repeat works great! The only problem is that the flattened proof is too large to generate, so we can't call `check_proof` on it, which checks the flattened version. Probably the solution is to disable `check_proof` for that benchmark.

mwillsey:
Is the lambda_function_repeat proof too large with shared subproofs? Or has that not been implemented yet?

oflatt:
It's not too large with shared subproofs, it's generated quickly. Its just too large without them.
For now I disabled that benchmark completely when proof generation is on, but I'd rather run it without checking the output proof.

One option is to write a smart proof-checker

src/macros.rs
oflatt: Also, this is awkward. In most cases, I would rather the PatternAst be a reference so it doesn't have to copy it. But custom appliers have to build their own ast often, so it can't be a reference there.

tests/lambda.rs
```diff
@@ -90,8 +97,16 @@ impl Analysis<Lambda> for LambdaAnalysis {
 
     fn modify(egraph: &mut EGraph, id: Id) {
         if let Some(c) = egraph[id].data.constant.clone() {
-            let const_id = egraph.add(c);
-            egraph.union(id, const_id);
+            let added = egraph.add_instantiation(&c.1, &Default::default());
```
oflatt, line 100: Since unions are delayed, analyses must add the pattern to the egraph before doing the union in order to call `union_with_justification`. Otherwise, the sub-terms in the pattern may not be unioned with the child yet.

oflatt:
Benchmarking results!

This branch compared to master with *explanation-generation turned off*
```name                          cycles       diff
diff_power_harder           35657999 +67.88506%
diff_power_simple            8327905 +56.81987%
integ_one                    1630352 +5.722366%
integ_part1                 12565989 +48.12221%
integ_part2                 56104386 +84.38759%
integ_part3                  5109986 +28.87924%
integ_sin                    1593095 +2.568898%
integ_x                      1750271 +8.634214%
math_associate_adds         83930599 +41.79915%
math_diff_different          1609494 +3.952402%
math_diff_ln                 1601508 +2.624456%
math_diff_same               1589239 +3.697645%
math_diff_simple1            5659196 +57.76904%
math_diff_simple2            4851992 +49.18185%
math_powers                  1777001 +7.553756%
math_simplify_add            3750376 +31.25842%
math_simplify_const          2491442 +26.52919%
math_simplify_factor        12666185 +50.32593%
math_simplify_root          15740714 +36.98765%
lambda_compose               6178888 +4.982307%
lambda_compose_many         23599773 +13.01355%
lambda_fib                 831980315 -4.062760%
lambda_if                    3040307 +8.758309%
lambda_if_elim               1654166 +3.284543%
lambda_if_simple              998206 +6.440110%
lambda_let_simple            1434397 +7.808221%
lambda_under                 1124396 +6.330186%
                                    +20.372964% mean```
                                    
                                    
This branch proof-production vs no proof-production, including generating the proof afterwards:
```diff_power_harder           45240302 +26.87280%
diff_power_simple           10138979 +21.74705%
integ_one                    1662628 +1.979695%
integ_part1                 16624151 +32.29481%
integ_part2                 75027092 +33.72768%
integ_part3                  6193361 +21.20113%
integ_sin                    1629783 +2.302939%
integ_x                      1796774 +2.656903%
math_associate_adds        130273400 +55.21562%
math_diff_different          1645303 +2.224861%
math_diff_ln                 1639374 +2.364397%
math_diff_same               1618268 +1.826598%
math_diff_simple1            6540076 +15.56546%
math_diff_simple2            5717959 +17.84766%
math_powers                  1836784 +3.364264%
math_simplify_add            4179892 +11.45261%
math_simplify_const          2661898 +6.841660%
math_simplify_factor        16676389 +31.66071%
math_simplify_root          19936513 +26.65571%
lambda_compose               6770424 +9.573503%
lambda_compose_many         28061991 +18.90789%
lambda_fib                 838312594 +0.761109%
lambda_if                    3264744 +7.382051%
lambda_if_elim               1751749 +5.899227%
lambda_if_simple             1007769 +0.958019%
lambda_let_simple            1504030 +4.854514%
lambda_under                 1141335 +1.506498%
                                    +12.188900% mean```

oflatt:
New benchmark results!
Here is explanations off compared to master. Overhead is from delaying unions- not delaying unions results in very small overhead.
```name                          cycles       diff
diff_power_harder           32469305 +51.94087%
diff_power_simple            7627469 +42.42304%
integ_one                    1648514 +4.960178%
integ_part1                 11586433 +35.34264%
integ_part2                 52549890 +69.89917%
integ_part3                  4819177 +20.33674%
integ_sin                    1636704 +3.478440%
integ_x                      1747543 +6.530932%
math_associate_adds         73407820 +23.19232%
math_diff_different          1650413 +4.682846%
math_diff_ln                 1645916 +3.586895%
math_diff_same               1631378 +4.511265%
math_diff_simple1            5260870 +45.17523%
math_diff_simple2            4511849 +37.20186%
math_powers                  1779901 +5.850718%
math_simplify_add            3403506 +17.61149%
math_simplify_const          2359417 +17.97409%
math_simplify_factor        11564720 +36.34551%
math_simplify_root          14240079 +23.18460%
lambda_compose               6016248 +0.989288%
lambda_compose_many         22120375 +5.222050%
lambda_fib                1586914124 -20.21437%
lambda_if                    2916077 +2.726695%
lambda_if_elim               1645739 +0.815843%
lambda_if_simple              992809 +3.211075%
lambda_let_simple            1412358 +3.857719%
lambda_under                 1122212 +3.615421%
                                    +13.792284% mean```

PR #119, opened 2021-09-07
Existance Explanations

oflatt:
Sometimes, it's unclear why an expression is represented in the egraph in the first place. In fact, these "magic terms" can sometimes be critical in finding an equality. Consider:
```
0
(Rewrite<= times-zero (* (/ 1 0) 0))
(Rewrite=> cancel-denominator 1)
```

`(* (/ 1 0) 0))` is a magic term, and we would like a proof of how it got to be in the egraph in the first place. I call these existence explanations, and this PR implements them. Whenever a term is added to the egraph, the `Explain` keeps track of where it came from. This could be either from a rule application, from a parent being inserted, or from congruence. Existance explanations simply follow the chain of pointers back to an original term manually added to the egraph.

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -59,7 +59,7 @@ pub struct EGraph<L: Language, N: Analysis<L>> {
     unionfind: UnionFind,
     #[cfg_attr(feature = "serde-1", serde(with = "vectorize"))]
     memo: HashMap<L, Id>,
-    to_union: Vec<(Id, Id, Option<Symbol>)>,
+    to_union: Vec<(Id, Id, Option<Symbol>, bool)>,
     pending: Vec<(L, Id)>,
     analysis_pending: IndexSet<(L, Id)>,
     #[cfg_attr(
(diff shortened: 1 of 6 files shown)


src/egraph.rs
```diff
@@ -285,15 +330,31 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
...
+                    let size_before = self.unionfind.size();
+                    let next_id = self.add(new_node);
+                    if self.unionfind.size() > size_before {
+                        new_node_q.push(true);
+                    } else {
+                        new_node_q.push(false);
+                    }
+                    if let Some(explain) = &mut self.explain {
+                        node.for_each(|child| {
+                            if new_node_q[usize::from(child)] {
```
oflatt, line 352: Only set existance to the parent if we created a new node

oflatt:
Interesting problem:
Right now, delaying unions means that analysis introduces new terms. For example, imagine (+ 1 (+ 2 3)) is in the egraph.
The analysis discovers (+ 2 3) = 5 and unions them, but the union is delayed. The analysis then discovers (+ 1 5) = 6, and the union is delayed. But the analysis had to insert (+ 1 5) *before* the union between (+ 2 3) and 5, thus adding a brand-new term the egraph hadn't seen before.

So this (+ 1 5) becomes another term added directly to the egraph, and we can't explain where it came from. I haven't thought of a good way to fix this, because explanations rely on inserting the left side of a term immediately.

sgpthomas:
Any progress on this PR? I'm interested in playing around with this. Are there any examples about how to use this with custom Searchers?

oflatt:
@sgpthomas We've decided to merge this! It'll be merged within a week.

PR #128, opened 2021-10-12
Add LP extractor

mwillsey:
This adds an Extractor that can do common subexpression elimination using linear programming. WIP


remysucre:
It can be worth trying the cycle breaking encoding in https://arxiv.org/pdf/1012.1802.pdf (grep for "There is an additional subtlety in the above encoding"). It introduces `n^2` variables where `n` is the number of enodes, but most of these variables will be inactive during ILP solving. It is complete, but will be slower than cycle-breaking a.la. tensat

Bastacyclop:
Hi! Do you know when this feature will be merged? Is it usable as-is, if I use this branch as my dependency? I'd like to play around with LP extraction.

mwillsey:
@Bastacyclop Yes this branch is usable and should be fully functional. I will try to merge it before the next release, and your feedback on using it would be very valuable!

Bastacyclop:
Hi @mwillsey. I started using this branch and there might be a useful feature missing: an easy way to add an "expression" with multiple roots back into an e-graph (a DFG really).
```rust
let (best, bestRoots) = LpExtractor::new(&runner.egraph, AstSize).solve_multiple(&roots[..]);                  
{ // used to print the outcome
  let mut g: EGraph<SymbolLang, ()> = Default::default();                                    
  g.add_expr(&best); // this function does not work, it would be great to have `add_dfg`                                                       
  g.dot().to_svg("/tmp/final.svg").unwrap();                                                 
}
```

Bastacyclop:
Also, I am getting a crash on:
https://github.com/egraphs-good/egg/blob/269c78636f91674da8c38838410d6fbdea08e099/src/lp_extract.rs#L186 
When extracting from the following graph with 3 roots:
image
Any idea why?

Bastacyclop:
Looks like your 'Canonicalize ids' commit was reverted somehow?

mwillsey:
@Bastacyclop great catch, I think I just fixed it.

PR #139, opened 2021-11-25
Fix broken link in Cargo.toml

krypt-n:
Hi,

I noticed that the repository link for egg on [crates.io](https://crates.io/crates/egg) leads to `https://github.com/egraph-good/egg` which does not seem to exist. The small change in this PR changes the URL in the cargo manifest to `https://github.com/egraphs-good/egg`. I assume that this will fix the link on creates.io when the next version of egg is released.

Feel free to close this PR and fix the link yourself, if there is some problem


mwillsey:
Good catch! Thanks!

PR #155, opened 2022-01-10
Fix Explanation Initialization to Share Subexpressions

oflatt:
This PR fixes https://github.com/egraphs-good/egg/issues/153, which points out that for large recexprs explanations are large and slow.
Previously, the explanations code failed to share subexpressions as `RecExpr` does when initializing the proof.


derekelkins:
There are still a few `println` statements in the commit that probably shouldn't be included.

mwillsey:
Good catch. I will fix shortly in a separate commit.

PR #162, opened 2022-01-20
Replace Mutex with RwLock in string interning code.

mmp:
When calling into egg from multiple threads, the `Mutex` used for `STRINGS` can become a bottleneck. (In my use case, I have a few hundred expressions that I'd like egg to simplify, so I do that using multiple threads.) In that context, I observed that `egg::util::intern` was a significant fraction of total runtime and that more time was spent in it than in any of the rest of egg's code(!). Here is what `perf` on Linux reported:

Screen Shot 2022-01-20 at 12 04 32 PM

(Note that is in the context of a larger application; just over half of total runtime is in egg but that 43% is with respect to total runtime.) Total runtime for my test is 4.04s user, 14.22s sys.

This PR replaces the `Mutex` is replaced with a reader-writer lock. `perf` reports roughly a 2x speedup:

Screen Shot 2022-01-20 at 12 12 45 PM

which is also reflected in the total runtime for my system going to 9.26s user, 0.21s sys.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,6 +1,6 @@
 use std::fmt;
 use std::str::FromStr;
-use std::sync::Mutex;
+use std::sync::RwLock;
 use symbolic_expressions::Sexp;
 
 use fmt::{Debug, Display, Formatter};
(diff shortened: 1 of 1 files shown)


mwillsey:
Great PR, I will accept after CI goes.

Out of curiosity, what were you doing that caused the interning to get called so frequently? My initial intention was that interning should only be called while setting things up.

mwillsey:
Ahh, that totally makes sense. If you are crossing FFI using strings, then you don't have much of a choice to parse!

PR #163, opened 2022-01-20
Shard the STRINGS in the intern cache for better parallel scalability

mmp:
(This is an improvement to the approach in #162 that is, however, probably not yet ready to merge. Posting it for discussion.)

The 20% of runtime in `egg::util::intern` with #162 is still pretty high. This PR instead shards the cache to reduce mutex contention in the presence of multithreading. With it, in the context of my application, it further reduces the percentage of time in `intern()` by a factor of 4 with a corresponding benefit to overall performance:

Screen Shot 2022-01-20 at 12 18 16 PM

As it stands currently, this PR unfortunately increases the size of `Symbol` to 8 bytes, which is probably not desirable. One option might be to hijack the high bits of the value in `Symbol` to store the shard index, if that would be acceptable.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,6 +1,6 @@
 use std::fmt;
 use std::str::FromStr;
-use std::sync::Mutex;
+use std::sync::RwLock;
 use symbolic_expressions::Sexp;
 
 use fmt::{Debug, Display, Formatter};
@@ -59,7 +59,14 @@ pub(crate) fn pretty_print(
     }
 }
 
-static STRINGS: Lazy<Mutex<IndexSet<&'static str>>> = Lazy::new(Default::default);
+const NUM_STRING_SHARDS: usize = 256;
+
+static STRINGS: Lazy<Vec<RwLock<IndexSet<&'static str>>>> = Lazy::new(|| {
+    let v: Vec<_> = std::iter::repeat_with(|| RwLock::new(Default::default()))
+        .take(NUM_STRING_SHARDS)
+        .collect();
+    v
+});
 
 /// An interned string.
 ///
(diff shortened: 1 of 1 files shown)


mmp:
(Another issue, perhaps, is the use of the same hash function both to choose a shard and then for inserting strings to the hash table. That could presumably potentially lead to an unexpectedly high number of hash collisions in a worst case scenario.)

mwillsey:
I agree, I'd rather not increase the size of `Symbol`. If this is a continued perf bottleneck, we could consider switching to a purpose-built interning solution (many exist in Rust, but I'm not familiar with all of them).

mwillsey:
Just leaving an idea here for now: You could "inline" < 3 byte strings into symbol pretty easily, which would require no synchronization for most variables (since most are very short).

Redefine `Symbol` as a `i32` newtype. If the sign bit is set, then just read the bottom three bits as the string. I guess the string would be 0 terminated, or you could stash the length in the top byte as well.

mmp:
Nice! I just got around to trying this with my workload; to my surprise it doesn't seem to give a performance benefit, FWIW.

mmp:
Ah, yes-for the example I was testing with, it happened that they were all >4 characters. Trying another one that has a mix of short ones and a few longer ones, I see a 10% reduction in overall wallclock time, so this is a definite win for that case, since it does a bunch of other stuff besides call into egg.

mwillsey:
Great! The latest commit on [tiny-symbols](https://github.com/egraphs-good/egg/tree/tiny-symbols) now implements both of these techniques, and it keeps the Symbol size to 4 bytes :). Barring any crazy results from you, I'll merge that and close this PR.

mwillsey:
Hi Matt! I've taken this idea and combined it with something inspired by [this post](https://matklad.github.io/2020/03/22/fast-simple-rust-interner.html) and made the [`symbol_table`](https://crates.io/crates/symbol_table) crate.

Commit ed1faf143b20ad93d0d266e88c94abf38ce6022c (and the upcoming 0.8 release) will use this for string interning, which should solve your perf issues once and for all!

PR #165, opened 2022-02-11
Instant Unions

oflatt:
[#115] Introduces explanations- proofs of equality between two terms in the egraph. To support the PR, egg was modified so that all unions were delayed until rebuild is called. This made explanations easier to implement because there is an up-to-date hashcons (memo for enodes in the egraph).

However, it turns out that carefully re-writing how enodes are added to the egraph avoids this issue. This PR restores egg's functionality to perform unions immediately. It also restores the rebuild algorithm to what it was before proofs were added. The core idea of the PR is to make sure that proofs have access to the correct uncanonicalized enode.

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -59,7 +59,6 @@ pub struct EGraph<L: Language, N: Analysis<L>> {
     unionfind: UnionFind,
     #[cfg_attr(feature = "serde-1", serde(with = "vectorize"))]
     memo: HashMap<L, Id>,
-    to_union: Vec<(Id, Id, Option<Symbol>, bool)>,
     pending: Vec<(L, Id)>,
     analysis_pending: IndexSet<(L, Id)>,
     #[cfg_attr(
@@ -108,7 +107,6 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
     pub fn new(analysis: N) -> Self {
         Self {
             analysis,
-            to_union: Default::default(),
             classes: Default::default(),
             unionfind: Default::default(),
             clean: false,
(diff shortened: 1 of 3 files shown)


src/egraph.rs
```diff
@@ -455,41 +483,70 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
...
-                parents: Default::default(),
-            };
 
-            // add this enode to the parent lists of its children
-            enode.for_each(|child| {
-                let tup = (enode.clone(), id);
-                self[child].parents.push(tup);
-            });
+            // now that we updated explanations, run the analysis for the new eclass
+            N::modify(self, id);
```
oflatt, line 520: @mwillsey it is key that we do modify after all the explanations stuff, so this version is different than the one we discussed.

src/egraph.rs
```diff
@@ -455,41 +483,70 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
...
+            self.clean = false;
+            id
+        }
+    }
 
-            // TODO is this needed?
-            self.pending.push((enode.clone(), id));
+    /// This function makes a new eclass in the egraph (but doesn't touch explanations)
+    fn make_new_eclass(&mut self, enode: L) -> Id {
```
oflatt, line 527: This is roughly the same as the old `add`

src/egraph.rs
```diff
@@ -519,88 +576,51 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
...
+    /// a `bool` indicating whether a union occured.
     pub fn union_instantiations(
         &mut self,
         from_pat: &PatternAst<L>,
         to_pat: &PatternAst<L>,
         subst: &Subst,
         rule_name: impl Into<Symbol>,
     ) -> (Id, bool) {
-        let id1 = self.add_instantiation(from_pat, subst);
+        let id1 = self.add_instantiation_internal(from_pat, subst);
```
oflatt, line 593: "Internal" functions now return uncanonicalized ids so we can track which enode explanations use.

PR #168, opened 2022-03-11
MultiPatterns

### LINKED ISSUE #39: Support multi-patterns for rewrite rules
**Description**

Currently the left hand side and the right hand side of a rewrite rule can only contain one subtree each, with the equivalent nodes being the two root nodes. It would be helpful to support more general cases where each side can contain multiple subtrees (potentially with shared nodes), and a rewrite rule can have multiple matched equivalent nodes.

**Example cases:**
1. (matmul ?input_1 ?input_2), (matmul ?input_1, ?input_3) => (split_0 (matmul ?input_1 (concat ?input_2 ?input_3))), (split_1 (matmul ?input_1 (concat ?input_2 ?input_3)))

Here split_0 (and split_1) refers to taking the first (and second) element from the split operation. matmul is matrix multiplication. There are two matched outputs: (matmul ?input_1 ?input_2) is equivalent to (split_0 (matmul ?input_1 (concat ?input_2 ?input_3))), and (matmul ?input_1, ?input_3) is equivalent to (split_1 (matmul ?input_1 (concat ?input_2 ?input_3))). They can't be specified as separate rules, since to apply (matmul ?input_1 ?input_2) => (split_0 (matmul ?input_1 (concat ?input_2 ?input_3))), we need to match ?input_3 in the egraph before as well (needs to also be matmul'ed with ?input_1). 

2. (relu ?input_1), (relu ?input_2) => (split_0 (relu (concat ?input_1 ?input_2))), (split_1 (relu (concat ?input_1 ?input_2)))

This case differs from above in that left hand side contains two separate subtrees, where case 1's LHS share common nodes between (matmul ?input_1 ?input_2) and (matmul ?input_1, ?input_3). Both would be useful.

**Some initial idea on how to achieve this**

Maybe we can still use the RecExpr to represent the patterns on both sides, but adding a mapping of the matched output pairs (instead of the two last elements of LHS and RHS). And when matching LHS (and applying RHS), we can keep a record on which indices of the RecExpr that hasn't been matched (and applied), so that we make sure all elements are matched (and applied) before we finish.

philzook58:
Multipatterns are multiple possibly disjoint patterns to seek out in a search process. They are distinct from guards in that they may bind new variables. This pull request attempts to implement somewhat efficient multipatterns in a backwards compatible way.
The pattern compiler can remain mostly as is. Compiler state about what patterns are where in the subst map needs to be retained between each single pattern compilation.
Multipatterns require changes to the abstract machine because scanning for roots of the 2nd pattern onwards becomes required.
Extending the interface to be able to give pattern names to the root of the pattern itself enables significant extra modelling power. Note that labelled multipatterns allows one to implement efficient manual scheduling of search even for single patterns via flattening or breaking the pattern trees up.

This PR is still a work in progress, but examples can be found in the [`datalog.rs`](https://github.com/philzook58/egg/blob/multipattern2/tests/datalog.rs) test. The syntax for creating multipatterns might change, but it looks this currently:
```rust
    let rules = vec![
        multi_rewrite!("base-case"; "?x = true = (edge ?a ?b)" => "?x = (path ?a ?b)"),
        multi_rewrite!("transitive"; "?x = true = (path ?a ?b) = (edge ?b ?c)" => "?x = (path ?a ?c)"),
    ];
```

Closes #39

### DIFF
--- a/src/multipattern.rs
+++ b/src/multipattern.rs
@@ -222,15 +222,15 @@ mod tests {
         assert_eq!(n_matches("?x = (f a a), ?x = (f a c)"), 0);
         assert_eq!(n_matches("?x = (f a b), ?x = (f a c)"), 1);
     }
-    
+
     #[test]
     fn unbound_rhs() {
         let mut egraph = EGraph::default();
         let _x = egraph.add_expr(&"(x)".parse().unwrap());
         let rules = vec![
-            // Rule creates y and z if they don't exist. Crashes with current parsing
+            // Rule creates y and z if they don't exist.
             rewrite!("rule1"; "?x = (x)" |- "?y = (y), ?y = (z)"),
-            // Can't fire. `y` and `z` don't already exist in egraph
+            // Can't fire without above rule. `y` and `z` don't already exist in egraph
             rewrite!("rule2"; "?x = (x), ?y = (y), ?z = (z)" |- "?y = (y), ?y = (z)"),
         ];
         let mut runner = Runner::default().with_egraph(egraph).run(&rules);
--- a/src/multipattern.rs
+++ b/src/multipattern.rs
@@ -171,14 +171,18 @@ impl<L: Language, A: Analysis<L>> Applier<L, A> for MultiPattern<L> {
     }
 
     fn vars(&self) -> Vec<Var> {
+        let mut bound_vars = HashSet::default();
         let mut vars = vec![];
-        // TODO are unbound binding vars allowed?
-        for (_v, pat) in &self.asts {
+        for (bv, pat) in &self.asts {
             for n in pat.as_ref() {
                 if let ENodeOrVar::Var(v) = n {
-                    vars.push(*v)
+                    // using vars that are already bound doesn't count
+                    if !bound_vars.contains(v) {
+                        vars.push(*v)
+                    }
                 }
             }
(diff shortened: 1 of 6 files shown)


philzook58:
In the right hand side, I think I can only express equality to a term bound in the lhs. In the Applier impl for MultiPattern, I would think one should allow creating new bindings only in the `?x = _` variable to give name to new terms constructed. See test `unbound_rhs`. I understand there is something suspicious about allowing new bindings in the rhs (only to give names to constructed terms), but I genuinely think this is useful and enhances expressivity.

src/multipattern.rs
```diff
@@ -0,0 +1,270 @@
...
+        egraph: &mut EGraph<L, A>,
+        matches: &[SearchMatches<L>],
+        _rule_name: Symbol,
+    ) -> Vec<Id> {
+        // TODO explanations?
+        // the ids returned are kinda garbage
+        let mut added = vec![];
+        for mat in matches {
+            for subst in &mat.substs {
+                let mut subst = subst.clone();
```
philzook58, line 156: This is an unfortunate clone, but I don't see an elegant way around it without changing interface of matches it be `&mut` or something.

mwillsey:
This was merged, but I guess the fast-forward make github not see it.

PR #176, opened 2022-05-05
Correct represented -> representation

tomjaguarpaw:
I assume this was the original intention, otherwise I'm not sure what this passage is trying to say!


mwillsey:
That’s actually supposed to say “a single represented term”! Good catch either way! If you update the PR, I’m happy to merge.

tomjaguarpaw:
OK done!

mwillsey:
Thanks!!

PR #184, opened 2022-06-10
Add an unsound method to easily merge to eclasses when explanations are enabled

oflatt:
Normally, unioning with explanations enabled involves using `union_instantiations`. This is a hassle because you need full `recexprs` or patterns that justify the union.

For e-class analysis demo purposes, it's easier to just let egg pick two "random" expressions in the e-class and use them to justify the union. That way you don't need to keep patterns around that justify your analysis. In the rare case that the reason that two eclasses are unioned does not depend on any specific term from either eclass, this method can be used.

One possible future of egg is that we return particular ids for particular terms in the egraph. Egg would return a particular id for your added expression, not the cannonical one for the eclass.  This gives you more power but makes it so that you have to be careful to canonicalize before you compare ids for e-class equality.


mwillsey:
This was merged in 0.9.0 in a different commit

PR #197, opened 2022-07-08
Add `search_with_limit`

yihozhang:
Currently, the `BackoffScheduler` will search each pattern, and if the match size is larger than the threshold, all the matches will be threw away and the rule will be banned. However, rules like AC will take an intractable amount of time and find ridiculously many matches, even when the threshold is small.

This PR tries to add an interface for `search_with_limit` and `search_eclass_with_limit`, while being compatible with the existing interface. With this PR, the e-matching machine will stop searching as soon as the number of matches hits the limit.

### DIFF
--- a/src/rewrite.rs
+++ b/src/rewrite.rs
@@ -84,6 +84,13 @@ impl<L: Language, N: Analysis<L>> Rewrite<L, N> {
         self.searcher.search(egraph)
     }
 
+    /// Call [`search_with_limit`] on the [`Searcher`].
+    ///
+    /// [`search_with_limit`]: Searcher::search_with_limit()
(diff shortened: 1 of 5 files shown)


src/rewrite.rs
```diff
@@ -128,7 +166,34 @@ where
 {
     /// Search one eclass, returning None if no matches can be found.
     /// This should not return a SearchMatches with no substs.
-    fn search_eclass(&self, egraph: &EGraph<L, N>, eclass: Id) -> Option<SearchMatches<L>>;
+    ///
+    /// Implementation of [`Searcher`] should implement one of
```
mwillsey, line 170: Hmm, I appreciate the effort to not break the API, but this could be a little confusing because the "default" behavior is an infinite loop! I think we should require `search_with_limit`.

PR #206, opened 2022-09-26
Fix the performance issue with math test suite

yihozhang:
Under an application-intensive scheduler like SimpleScheduler, math could take 100% of its time doing the cloning in its analysis. See the screenshot below:
Screen Shot 2022-09-26 at 2 28 18 PM


mwillsey:
Nice find!

PR #210, opened 2022-10-23
Handle overflow in proof size

oflatt:
This PR handles the overflow in proof size estimates by capping them at `usize::MAX`.

I considered using floating-point, but it's harder to reason about floating-point numbers and they also do not implement `Ord` which is a problem for the heap implementation.

Another thing to consider would be using a bignum, but I'm not sure if it matters.

### DIFF
--- a/src/explain.rs
+++ b/src/explain.rs
@@ -1472,18 +1472,19 @@ impl<L: Language> Explain<L> {
         );
 
         // calculate distance to find upper bound
-        b.checked_add(c)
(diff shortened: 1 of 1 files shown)


mwillsey:
This looks good! It seems like this may be a case where you should prefer [`saturating_add`](https://doc.rust-lang.org/stable/std/primitive.usize.html#method.saturating_add) over `checked_add`.

oflatt:
Fixed! Ready to merge

PR #243, opened 2023-04-03
Optimize path compression of union-find

QuarticCat:
Before:

```console
$ hyperfine --warmup=3 'cargo test --test lambda --release -- --nocapture --test --test-threads=1'
Benchmark 1: cargo test --test lambda --release -- --nocapture --test --test-threads=1
  Time (mean ± σ):     870.4 ms ±  22.7 ms    [User: 845.0 ms, System: 23.2 ms]
  Range (min … max):   841.0 ms … 912.3 ms    10 runs
```

After:

```console
$ hyperfine --warmup=3 'cargo test --test lambda --release -- --nocapture --test --test-threads=1'
Benchmark 1: cargo test --test lambda --release -- --nocapture --test --test-threads=1
  Time (mean ± σ):     848.1 ms ±  16.2 ms    [User: 823.6 ms, System: 22.3 ms]
  Range (min … max):   829.7 ms … 874.2 ms    10 runs
```

If we can eliminate the boundary check of index operations, the performance could be a bit faster.

### DIFF
--- a/src/unionfind.rs
+++ b/src/unionfind.rs
@@ -33,13 +33,11 @@ impl UnionFind {
         current
     }
 
-    pub fn find_mut(&mut self, mut current: Id) -> Id {
-        while current != self.parent(current) {
-            let grandparent = self.parent(self.parent(current));
-            *self.parent_mut(current) = grandparent;
-            current = grandparent;
+    pub fn find_mut(&mut self, current: Id) -> Id {
+        if current != self.parent(current) {
+            *self.parent_mut(current) = self.find_mut(self.parent(current));
         }
-        current
+        self.parent(current)
     }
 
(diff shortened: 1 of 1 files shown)


QuarticCat:
The tests `pattern::tests::simple_match` and `explain::tests::simple_explain` contain hard-coded paths to the building target dir. If someone changes their target dir settings (like me), these two tests will break. Would you like to fix it? I can add a commit to this thread.

QuarticCat:
After I updated my benchmarking environment, the speedup is negligible. So I gonna close this PR. The bench script can be found here https://github.com/QuarticCat/dotfiles/blob/main/zsh/scripts/bench-mode.zsh

PR #250, opened 2023-04-24
Use a queue for pending_analysis

### LINKED ISSUE #239: Pathological rebuild performance
Consider an analysis that is overwhelmingly likely to propagate updates throughout the e-graph. In my use-case, I was computing `AstSize` in my `Analysis`, as I needed it for some heuristics. The below example models this by always returning `DidMerge(true, true)`.

```rust
use egg::*;

// Runtime is extremely sensitive to this argument
const SIZE: usize = 14;

type EGraph = egg::EGraph<SymbolLang, MyAnalysis>;

struct MyAnalysis;

impl Analysis<SymbolLang> for MyAnalysis {
    type Data = ();

    fn make(_: &EGraph, _: &SymbolLang) {}
    fn merge(&mut self, _: &mut (), _: ()) -> DidMerge {
        DidMerge(true, true)
    }
}

fn wrap(egraph: &mut EGraph, mut v: Id) {
    for _ in 0..SIZE {
        let f = egraph.add(SymbolLang { op: "f".into(), children: vec![v] });
        let g = egraph.add(SymbolLang { op: "g".into(), children: vec![v] });
        egraph.union(f, g);
        v = f;
    }
}

fn main() {
    env_logger::init();

    let mut egraph = EGraph::new(MyAnalysis);
    let x = egraph.add(SymbolLang { op: "x".into(), children: vec![] });
    let y = egraph.add(SymbolLang { op: "y".into(), children: vec![] });
    wrap(&mut egraph, x);
    wrap(&mut egraph, y);
    egraph.rebuild();

    egraph.union(x, y);
    egraph.rebuild();
}
```

Running this with `RUST_LOG=egg=info cargo r --release` prints:

```
[2023-02-21T05:20:57Z INFO  egg::egraph] REBUILT! in 0.000s
      Old: hc size 58, eclasses: 30
      New: hc size 58, eclasses: 30
      unions: 0, trimmed nodes: 0
[2023-02-21T05:21:15Z INFO  egg::egraph] REBUILT! in 18.454s
      Old: hc size 58, eclasses: 29
      New: hc size 58, eclasses: 15
      unions: 14, trimmed nodes: 28
```

The issue is that `process_unions` iterates through `analysis_pending` in a random order, instead of bottom-up. Performing the latter accurately might be non-trivial, but while I haven't thought through this fully, I believe a very good approximation, that would recover the optimal asymptotic complexity, is to count how many times the analysis data of each e-node has been handled during the current rebuild, and in each iteration pick the one where this count is lowest (with something like `BinaryHeap`).

mwillsey:
Fixes #239

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -1046,6 +1046,8 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
     fn process_unions(&mut self) -> usize {
         let mut n_unions = 0;
 
+        let mut tmp_analysis_pending = IndexSet::default();
+
         while !self.pending.is_empty() || !self.analysis_pending.is_empty() {
             while let Some((mut node, class)) = self.pending.pop() {
                 node.update_children(|id| self.find_mut(id));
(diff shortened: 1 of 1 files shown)


Bastacyclop:
Looks good to me except that there can be duplicate elements between `analysis_pending` and `tmp_analysis_pending`. I'd personally rather combine `IndexSet` (or `HashSet`) with a `Vec` as proposed in #239 , to get real unique queue behaviour. Happy to make a PR on Friday with that alternative solution, it's really not a lot of code.

What do you think @meithecatte ?

PR #259, opened 2023-05-17
More fixes for large proofs (overflows)

oflatt:


mwillsey:
Happy to merge, but have you seen this? Could be a cleaner alternative: https://docs.rs/saturating/0.1.0/saturating/index.html

oflatt:
Oh I haven't seen that, looks like a better solution

PR #266, opened 2023-08-18
Add class size based scheduler

remysucre:
This is similar to the Backoff Scheduler, but bans rewrites based on the class size. Specifically, when a class grows past a limit, we stop applying rewrites to it for a few iterations. 

This is more fine-grained than the Backoff Scheduler, which bans a rule for all classes when that rule explodes some class. Scheduling based on class size is also closer to how we do prove by hand: if we have rewritten a certain subterm too many times, we'll move on to some other subterms. For example, if we've applied AC to a certain class too many times, we'll still apply AC, but to other classes.

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -761,14 +761,14 @@ impl Default for EScheduler {
 impl EScheduler {
     /// Set the initial size limit after which a class will stop growing.
     /// Default: 10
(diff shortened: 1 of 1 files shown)


mwillsey:
I think we'll close this for now, as it's implementable as an external library (as demonstrated by this PR).

PR #270, opened 2023-09-18
allow per-language discriminants

bjchambers:
For languages like `SymbolLang` that don't have an outer-most enum this allows defining a discriminant and reducing the set of nodes considered for matching. This allows each `Language` to determine the type to use for the discriminant and how to compute it.

For instance:

1. This may be a `std::mem::Discriminant<Self>` if the type is an enum
2. This may be a `u64` if it is easy to hash the "operation"
3. This may be `Symbol` for something like SymbolLang (or a discriminant of the operation, in the case of something like SymbolLang but using an enum for the symbol)
4. This may be something a bit richer for cases like `ENodeOrVar`.

This also allows languages to trade-off how rich the discriminant is with how efficient it is to compute / compare.

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -42,7 +42,6 @@ fn for_each_matching_node<L, D>(
 where
     L: Language,
 {
-    #[allow(enum_intrinsics_non_enums)]
     if eclass.nodes.len() < 50 {
         eclass
             .nodes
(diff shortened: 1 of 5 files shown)


mwillsey:
This looks great! Could you please document the associated type as well as in the `define_language` macro that it's using the enum discriminant.

bjchambers:
> This looks great! Could you please document the associated type as well as in the `define_language` macro that it's using the enum discriminant.

Absolutely. Wanted to get it out and make sure it seemed reasonable before polishing it too much. Will do so now!

src/machine.rs
```diff
@@ -53,17 +52,17 @@ where
         debug_assert!(node.all(|id| id == Id::from(0)));
         debug_assert!(eclass.nodes.windows(2).all(|w| w[0] < w[1]));
         let mut start = eclass.nodes.binary_search(node).unwrap_or_else(|i| i);
-        let discrim = std::mem::discriminant(node);
+        let discrim = node.discriminant();
         while start > 0 {
-            if std::mem::discriminant(&eclass.nodes[start - 1]) == discrim {
+            if eclass.nodes[start - 1].discriminant() == discrim {
```
bjchambers, line 57: I wasn't sure about this check here. Why is it expected that nodes with equal discriminant will be adjacent? It doesn't seem like the nodes are sorted by discriminant (the Discriminant doesn't implement Ord).

bjchambers:
Comments added.

Issue #276, opened 2023-10-07
Mutable access to the EGraph in `Analysis::make`


cole-k:
Can we change the signature of `Analysis::make` as follows? I need to access my EGraph analysis mutably when I make an EClass analysis.

```diff
-    fn make(egraph: &EGraph<L, Self>, enode: &L) -> Self::Data {
+    fn make(egraph: &mut EGraph<L, Self>, enode: &L) -> Self::Data {
```

It seems to not have any problems locally.

The change required: https://github.com/egraphs-good/egg/compare/main...cole-k:egg:mutable-egraph-in-analysis-make

cole-k:
I realize now that this may have been intentional to prevent screwing with the egraph during the analysis. My quandary is that I need mutable access to the EGraph analysis, any suggestions on how to resolve this? I don't mind writing some extra code.

mwillsey:
Could you use a refcell?

cole-k:
Yeah I think that works. I'll close this then.

PR #277, opened 2023-10-09
Make Analysis::make take in the E-graph mutably

imbrem:
I have a use case where I want to tag each value in the E-graph with type information, which itself should be stored either in the E-graph or in another E-graph stored in the Analysis. This requires me to mutate the E-graph or at least the Analysis from inside `Analysis::make` (the alternative would be to wrap a nested E-graph in the analysis in `RefCell` or `RwLock`); thankfully, just making the E-graph reference passed in mutable is enough to get this to work with only minimal changes (sticking an `&mut` on implementations of `Analysis`) to anything else. Unfortunately, this is still a breaking change, so not sure if you guys would be interested in it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changes
 
 ## [Unreleased] - ReleaseDate
+- Change the API of `make` to have mutable access to the e-graph for some [advanced uses cases](https://github.com/egraphs-good/egg/pull/277).
 
 
 ## [0.9.5] - 2023-06-29
--- a/src/language.rs
+++ b/src/language.rs
@@ -691,11 +691,14 @@ pub trait Analysis<L: Language>: Sized {
     /// The per-[`EClass`] data for this analysis.
     type Data: Debug;
 
-    /// Makes a new [`Analysis`] for a given enode
-    /// [`Analysis`].
+    /// Makes a new [`Analysis`] data for a given e-node.
     ///
-    /// Note that the mutable `egraph` parameter allows creating new nodes, for which `make` will be called recursively; this may lead to an infinite loop if not careful!
-    /// 
+    /// Note the mutable `egraph` parameter: this is needed for some
+    /// advanced use cases, but most use cases will not need to mutate
+    /// the e-graph in any way.
+    /// It is **not** `make`'s responsiblity to insert the e-node;
+    /// the e-node is "being inserted" when this function is called.
+    /// Doing so will create an infinite loop.
     fn make(egraph: &mut EGraph<L, Self>, enode: &L) -> Self::Data;
 
     /// An optional hook that allows inspection before a [`union`] occurs.
--- a/src/language.rs
+++ b/src/language.rs
@@ -651,7 +651,7 @@ impl Analysis<SimpleMath> for ConstantFolding {
         egg::merge_max(to, from)
     }
 
-    fn make(egraph: &EGraph<SimpleMath, Self>, enode: &SimpleMath) -> Self::Data {
+    fn make(egraph: &mut EGraph<SimpleMath, Self>, enode: &SimpleMath) -> Self::Data {
         let x = |i: &Id| egraph[*i].data;
         match enode {
             SimpleMath::Num(n) => Some(*n),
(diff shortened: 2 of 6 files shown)


yihozhang:
Does the solution proposed in #276 work for your setting?

imbrem:
Not particularly well, since I'd like my type to be `Sync` since once the e-graph rewriting is done many threads might inspect it

mwillsey:
@imbrem I think wrapping your Analysis type in a mutex should work, right? I'm a little hesitant to break the API if we don't need to.

imbrem:
@mwillsey well we'd need an `RwLock` for acceptable performance, but even that introduces overhead, not to mention hassles like e.g. manual `Clone` implementation, and to a degree having "too much freedom." Besides, it seems inconsistent with the rest of the API, which is passed a mutable `Analysis` (e.g., `merge`), to have `make` not receive a mutable analysis, especially since it doesn't break any of the test cases.

One potential non-breaking option is to make a new `AnalysisMut` trait and have a blanket impl for things which implement `Analysis`, but I don't think this is very clean.

mwillsey:
Sure, that all makes sense. I still don't have a great understanding of your use case though. It seems like perhaps a "hook" of some kind on e-node insertion could suffice (and would not break the API).

imbrem:
A simplified version of my use case is as follows: there's some kind of map in my analysis, and I want every E-node analysis update to be able to not only look at what's currently in the map, but also to be able to change the map so everything points to the right value. Doing this on insertion would not allow changing the value in the map; for my _current_ needs this is possible to work around, but again it requires a bit of "yoga" (I need to, for example, have a guarantee that I did the insertion in the hook which runs *before* `make`, or trying to run the `make` function will panic), and it rules out some potential future designs as well.

mwillsey:
To make that a big more concrete: we could add something like
```rust
fn enode_just_added(egraph: &mut EGraph<L, Self>, enode: &L, data: Self::Data);
```
or perhaps a hook that runs before insertion. Would that suffice? I'm trying to disentangle what you need from `make`, since there are other uses of `make` outside of e-node insertion (like [this one in rebuilding](https://github.com/egraphs-good/egg/blob/6340768fbfc1665d5a1b28c39f385277c71b11c9/src/egraph.rs#L1065)).

imbrem:
The issue is that I need to mutate the EGraph, or specifically the analysis, to _generate_ the Data in the first place, since it's an index into a map which _may_ not yet exist.

mwillsey:
Ok, I see your point. @oflatt @yihozhang any thoughts?

In the documentation, it would be nice to clarify that it is *not* `make`'s responsibility to insert the e-node; that has already been done.

(ps: your use case sounds like it might be nicer in [egglog](https://github.com/egraphs-good/egglog); you could express this natively with a `typeof` function or something. caveat: egglog is not quite as polished of a tool as egg right now)

yihozhang:
The example makes sense to me. I think to be consistent with other functions like `merge`, what we really want is mutable access to `Analysis` and immutable access to `EGraph` at the same time. But there's no trivial way of separating them.

imbrem:
@yihozhang well, actually, mutable access to the E-graph as well allows us to be even _more_ flexible so long as we don't trigger an infinite loop, as in this case we can do dependent-type-style things (e.g. generate Ids for enodes while making data for the analysis). This should be fine so long as the enodes passed in are of a different kind which will not lead to a recursive call, which shouldn't be _too_ hard to guarantee.

yihozhang:
Yeah, though I think we want E-class analysis to strike the right balance between being flexible and being a good, sound abstraction. In general, however, unrestrained access to E-graphs seems ad-hoc and dangerous (as you said, now you need to reason about how the internals of egg work). In the egg paper, we just require an e-class analysis to be a lattice and everything just works. 

Here is another proposal to `make`'s interface:
before:
```
    fn make(egraph: &EGraph<L, Self>, enode: &L) -> Self::Data;
```
after:
```
    fn make(&mut self, enode: &L, children: &[Data]) -> Self::Data;
```

This is "more faithful" to the idea that the e-class analysis of a parent is the composition of the e-class analyses of its children, and it seems this works for your use case. This seems to be a bigger change, though.

Either way, I'm fine with the current change (as well as my proposed change).

imbrem:
@yihozhang I still favor the proposed change, since the requirement is not only relatively minor but also only relevant to advanced users. There could even be a second make method implementing my API, with a default implementation in terms of your simplified version; but again it is useful to be able to look at the EGraph in make so... eh, I say just provide the flexibility if it's possible without any real additional complexity right?

mwillsey:
Ok, I think this change is worth the cost; it definitely opens up some more opportunities to do things in the e-graph that would be easier in egglog, but it worth supporting them here in egg I think.

Issue #282, opened 2023-11-25
Explanations using `add_internal` and `union_trusted`


dewert99:
The `EGraph::union_trusted` method currently has the documentation:

> Unions two e-classes, using a given reason to justify it.
>
> Unlike union_instantiations, this function picks arbitrary representatives from either e-class. When possible, use [union_instantiations](https://docs.rs/egg/latest/egg/struct.EGraph.html#method.union_instantiations), since that ensures that the proof rewrites between the terms you are actually proving equivalent.

As far as I can tell this "arbitrariness" comes from the fact that publicly exposed methods for getting `Id`s from the `EGraph` (`add`, `add_expr`, `add_instantiation`, `find`, etc.) canonicalize the `Id`s they return. Using `union_trusted` with the `Id`s returned by the internal variants of these methods seems to produce expected explanations.
eg.
```rust
    #[test]
    fn simple_explain_union_trusted() {
        crate::init_logger();
        let mut egraph = EGraph::new(()).with_explanations_enabled();

        let a = egraph.add_internal(SymbolLang::leaf("a"));
        let b = egraph.add_internal(SymbolLang::leaf("b"));
        let c = egraph.add_internal(SymbolLang::leaf("c"));
        let d = egraph.add_internal(SymbolLang::leaf("d"));
        egraph.union_trusted(a, b, "a=b");
        egraph.rebuild();
        let fa = egraph.add_internal(SymbolLang::new("f", vec![a]));
        let fb = egraph.add_internal(SymbolLang::new("f", vec![b]));
        egraph.union_trusted(c, fa, "c=fa");
        egraph.union_trusted(d, fb, "d=fb");
        egraph.rebuild();
        let exp = egraph.explain_equivalence(&"c".parse().unwrap(), &"d".parse().unwrap());
        println!("{:#}", exp.get_sexp());
    }
```
produces the explanation
```
(Explanation c (Rewrite=> c=fa (f a)) (f (Explanation a (Rewrite=> a=b b))) (Rewrite<= d=fb d))
```
but using `egraph.add` instead of `egraph.add_internal` produces
```
(Explanation c (Rewrite=> c=fa (f a)) (Rewrite<= d=fb d))
```

Is my understanding correct? Is the behaviour expected to be stable? If so, would it be worth making the internal variants of these functions public (and changing their names), and updating the documentation of `union_trusted`? 

[...]

PR #283, opened 2023-11-27
Add public functions for working with uncanonicalized Ids

dewert99:
Implements the changes described in #282

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -342,6 +342,15 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
         }
     }
 
+    /// Like [`id_to_expr`](EGraph::id_to_expr) but only goes one layer deep
+    pub fn id_to_node(&self, id: Id) -> &L {
+        if let Some(explain) = &self.explain {
+            explain.node(id)
+        } else {
+            panic!("Use runner.with_explanations_enabled() or egraph.with_explanations_enabled() before running to get unique expressions per id");
+        }
+    }
+
     /// Like [`id_to_expr`](EGraph::id_to_expr), but creates a pattern instead of a term.
     /// When an eclass listed in the given substitutions is found, it creates a variable.
     /// It also adds this variable and the corresponding Id value to the resulting [`Subst`]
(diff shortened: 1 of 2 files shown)


dewert99:
Fixes #282

src/egraph.rs
```diff
@@ -428,7 +445,13 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
     /// Note that this function can be called again to explain any intermediate terms
     /// used in the output [`Explanation`].
     pub fn explain_existance(&mut self, expr: &RecExpr<L>) -> Explanation<L> {
-        let id = self.add_expr_internal(expr);
+        let id = self.add_expr_uncanonical(expr);
+        self.explain_existance_id(id)
+    }
+
+    /// Equivalent to calling [`explain_existance`](EGraph::explain_existance)`(`[`id_to_expr`](EGraph::id_to_expr)`(id))`
+    /// but more efficient
```
oflatt, line 453: Good documentation, but I think it needs a little more explanation.
Perhaps something like a warning that you need uncanonical ids, and how you might get those uncanonical ids in the first place.

src/egraph.rs
```diff
@@ -583,11 +609,16 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
...
+        let id = self.add_instantiation_noncanonical(pat, subst);
         self.find(id)
     }
 
-    fn add_instantiation_internal(&mut self, pat: &PatternAst<L>, subst: &Subst) -> Id {
+    /// Similar to [`add_instantiation`](EGraph::add_instantiation) but the `Id` returned may not be
+    /// canonical
+    ///
+    /// Like [`add_uncanonical`](EGraph::add_uncanonical), when explanations are enabled calling
+    /// [`id_to_expr`](EGraph::id_to_expr) on this `Id` will correspond to the parameter `enode`
```
oflatt, line 620: Confusing because there is no `enode` parameter. There is a pattern and subst.

src/egraph.rs
```diff
@@ -833,8 +880,6 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
     ///
     /// When explanations are enabled, this function behaves like [`EGraph::union_trusted`],
     ///  and it lists the call site as the proof reason.
-    /// You should prefer [`union_instantiations`](EGraph::union_instantiations) when
```
oflatt, line 836: Don't remove this- `union_instantiations` is still preferred since it is the least confusing. 
However, you could add a note that alternatively, use `union_trusted` with uncanonical ids from `add_uncanonical`.

oflatt (changes requested):
I think this is a good change and a helpful API, but in general we should put more warnings everywhere. Most egg users shouldn't user or don't want uncanonical ids.

oflatt (approved):
Great job! Thanks for taking the time to improve egg

PR #284, opened 2023-11-28
Optimised the `classes` field of `EGraph` to avoid extra hashing

dewert99:
I noticed that the `classes` field of an `EGraph` is currently represented using a `HashMap<Id, EClass>` (or possibly an `IndexMap`). Since the set of possible `Id`s form a continuous range and the `union_find` field already contains an element for each `Id`, I thought I could try to use it like an "index" for an `IndexMap`, this works especially well since it currently maps non-canonical `Id`s to there parent, and we would need it to map canonical `Id`s to the index into the class list, so each element of the `union_find` would only need to have an index into another list and a tag of whether or not it's canonical. I decided to use one bit of a u32 as the tag so that the `union_find` would stay the same size.

Hopefully, this will improve performance (do you have a setup for reliably benchmarking two commits and comparing them?)

The main downside to this change is that `EGraph`s can now only support 2^31 `Id`s instead of 2^32 and that we now rely more on the `EClass`s `id` field so if a user changes it (this is possible since it is public) it would likely break the `EGraph` (if it didn't already).

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -73,7 +73,7 @@ pub struct EGraph<L: Language, N: Analysis<L>> {
             deserialize = "N::Data: for<'a> Deserialize<'a>",
         ))
     )]
-    pub(crate) classes: HashMap<Id, EClass<L, N::Data>>,
+    pub(crate) classes: Vec<EClass<L, N::Data>>,
     #[cfg_attr(feature = "serde-1", serde(skip))]
     #[cfg_attr(feature = "serde-1", serde(default = "default_classes_by_op"))]
     pub(crate) classes_by_op: HashMap<L::Discriminant, HashSet<Id>>,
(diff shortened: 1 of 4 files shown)


mwillsey:
Interesting idea, seems like it's probably faster but I'd want to make sure before merging something like this with non-trivial complexity.

You can run the built-in test suite and control various knobs as seen in the [test.rs file](https://github.com/egraphs-good/egg/blob/main/src/test.rs#L45). Something like
```shell
EGG_BENCH_CSV=out.csv cargo test --release --test=math -- --test-threads=1
```

You need test threads to be 1 to avoid concurrent writes to the CSV. You can do the same for the `lambda` benchmark suite.

Or you could try the e-matching benchmark suites:

```sh
EGG_ITER_LIMIT=5 cargo test --release -- --nocapture bench
```

Try tuning that limit up to make a more "significant" benchmark run.

mwillsey:
Nice work! Yes, this is sort of what I expected. I think the cost of this change isn't worth the complexity, since performance is so dominated by other factors. So I think we should close this for now unless/until a need is demonstrated.

Issue #219, opened 2022-12-05
Add an example that **doesn't use** the define_language macro?


sheredom:
It'd be great if one of the tests (that you point to as examples of how to setup egg) didn't use the define_language macro - as I actually don't have a clue how to hook up egg to my toy compiler IR without having some sort of textual representation! I'm sure there _must_ be a way for me to hook up my toy IR to egg, I'm just not sure how!

samuelgruetter:
I was missing such an example too! I eventually figured something out: I always use the predefined `SymbolLang`, and dynamically construct rewrites using `Rewrite::new`, and `.parse::<Pattern<SymbolLang>>().unwrap()` to construct a pattern from a string.
Not sure if that's what you're looking for, but here's some code I wrote:
https://github.com/samuelgruetter/egg/blob/344678d9465e8ac6ce76af132adf72caadf3c303/src/bin/coquetier.rs#L42

mwillsey:
You can use any type you want! You just have to implement the [`Language`](https://docs.rs/egg/latest/egg/trait.Language.html) trait (and [`FromOp`](https://docs.rs/egg/latest/egg/trait.FromOp.html) if you want parsing and rust's `Display` if you want printing).

sheredom:
How do I write `Rewrite`'s without using a textual format for them? Couldn't work that out either!

mwillsey:
It's quite difficult to make `Rewrite`s without parsing. Note that you can implement `FromOp` yourself to get the ability to parse patterns. If you don't want to implement `FromOp`, then it you can manually create a [`PatternAst`](https://docs.rs/egg/latest/egg/type.PatternAst.html) can then you can call [`Pattern::new`](https://docs.rs/egg/latest/egg/struct.Pattern.html#method.new) on that. Then you can call [`Rewrite::new`](https://docs.rs/egg/latest/egg/struct.Rewrite.html#method.new) on a two patterns.

sheredom:
Ok I'll try that ta!

valarauca:
Seconding this request.

Trying to incorporate this crate into my existing DLS feels daunting to a near impossible task.

mwillsey:
You might find the [implementation](https://docs.rs/egg/latest/src/egg/language.rs.html#828-840) of `Language` for the built-in `SymbolLang` somewhat helpful.

valarauca:
I found that code sample, it wasn't.

mwillsey:
[...]

PR #287, opened 2023-12-11
An approach to a string-free API

robert-chiniquy:
This change is to support a string-free API to create rewrites as discussed in https://github.com/egraphs-good/egg/issues/219

I do need this functionality to use egg with my own IR, so it would be great if it landed. 

However, if supporting a string-free API is better done using some other approach, I'm happy to do that instead. Just doing this to illustrate the idea.

## Changes

- Make `ENodeOrVar::Var(V)`, adding a generic V in place of `Var(GlobalSymbol)`
- all the associated fallout of adding a generic across the type signatures in the codebase (Note though that no existing test needed to be changed to support this, as `V` defaults to `Var`)
- Add a "No Strings" test to illustrate and prove the API

### Trait constraints generally 

Trait constraints generally are slapdash and could be reduced, I'll put in the work to clean this up if this approach to the change is approved.

### Adds a constraint to struct definition

I'm not happy about this specific change, hoping it is not needed:

```rs
pub struct SearchMatches<'a, L: Language, V = Var>
where
    V: Clone,
```

### `Explanation::check_rewrite` type inference

For some reason, this caused type inference for `N` to fail in a few spots for`Explanation::check_rewrite()`, so it is explicitly specified here.

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -626,7 +626,10 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
     /// Like [`add_uncanonical`](EGraph::add_uncanonical), when explanations are enabled calling
     /// Calling [`id_to_expr`](EGraph::id_to_expr) on this `Id` return an corrispond to the
     /// instantiation of the pattern
-    fn add_instantiation_noncanonical(&mut self, pat: &PatternAst<L>, subst: &Subst) -> Id {
+    fn add_instantiation_noncanonical<V>(&mut self, pat: &PatternAst<L, V>, subst: &Subst<V>) -> Id
+    where
+        V: std::fmt::Debug + Clone + std::hash::Hash + PartialOrd + Ord + Copy + std::fmt::Display,
+    {
         let nodes = pat.as_ref();
         let mut new_ids = Vec::with_capacity(nodes.len());
         let mut new_node_q = Vec::with_capacity(nodes.len());
@@ -850,13 +853,16 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
     ///
     /// Returns the id of the new eclass, along with
     /// a `bool` indicating whether a union occured.
-    pub fn union_instantiations(
+    pub fn union_instantiations<V>(
         &mut self,
-        from_pat: &PatternAst<L>,
-        to_pat: &PatternAst<L>,
-        subst: &Subst,
+        from_pat: &PatternAst<L, V>,
+        to_pat: &PatternAst<L, V>,
+        subst: &Subst<V>,
         rule_name: impl Into<Symbol>,
-    ) -> (Id, bool) {
+    ) -> (Id, bool)
+    where
+        V: std::fmt::Debug + Clone + std::hash::Hash + PartialOrd + Ord + Copy + std::fmt::Display,
+    {
         let id1 = self.add_instantiation_noncanonical(from_pat, subst);
         let size_before = self.unionfind.size();
         let id2 = self.add_instantiation_noncanonical(to_pat, subst);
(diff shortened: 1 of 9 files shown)


robert-chiniquy:
Oh, one more thing, just thought I'd highlight the actual API here so people don't have to dig through the diff, it looks like this:

```rs

define_language! {
    enum NoStrings {
        Num(i32),
        "🦑" = Squid([Id; 2]),
        "🐝" = Bee(Id),
        Symbol(Symbol),
    }
}

fn make_rules() -> Vec<Rewrite<NoStrings, (), i32>> {
    let searcher: Pattern<NoStrings, _> = Pattern::new(RecExpr::from(vec![
        ENodeOrVar::Var(12 as i32),
        ENodeOrVar::Var(13 as i32),
        ENodeOrVar::ENode(NoStrings::Squid([0.into(), 1.into()])),
    ]));
    let applier: Pattern<NoStrings, _> = Pattern::new(RecExpr::from(vec![
        ENodeOrVar::Var(13 as i32),
        ENodeOrVar::ENode(NoStrings::Bee(0.into())),
    ]));
    vec![crate::Rewrite::new("squid to bee".to_string(), searcher, applier).unwrap()]
}

#[test]
fn simple_tests() {
    assert_eq!(
        simplify(RecExpr::from(vec![
            NoStrings::Num(77 as i32),
            NoStrings::Num(88 as i32),
            NoStrings::Squid([0.into(), 1.into()]),
        ])),
        RecExpr::from(vec![NoStrings::Num(88 as i32), NoStrings::Bee(0.into()),])
    );
}

```

dewert99:
I'm not sure if this is relevant to your use case @robert-chiniquy but in order to have a truly string-free API I think the e-graph to be parameterized by the rule names used in explanations as they currently need to be `GlobalSymbol`s

mwillsey:
@robert-chiniquy, what type will you be plugging in for `V` in your application? Is it "integer-like"?

robert-chiniquy:
Yep, i am currently. I feel like that should work for at least almost all use cases.

mwillsey:
Ok. I'm really hesitant to make changes to egg's user-facing API at this point, since most of the egg folks (including me) are spending most of our time on other projects at the moment. Are you worried about the performance overhead of serializing your `V` types to a string? I think it shouldn't be too bad, especially since it only has to be done upfront for the rewrite rules.

robert-chiniquy:
Right, I understand. I had kind of thought the way I did this (with default `Var` for the generic `V` in all sites) would be sufficient to not change the user-facing API for existing users. I understand your concern there though if I missed something there. 

Just like anyone, I care about performance— but yeah, if it were just serializing to string as the cost, the compute performance hit on that wouldn't be that bad. 

Aren't I right though that if I serialize to string to use `GlobalSymbol`, then my strings are permanently interned? My use case involves long-running processes (and memory-constrained systems); when using a `usize`, as the value is there to uniquely represent a quantification, no additional external storage for the usize outside of the egraph is required. If each number I use is serialized to string and permanently interned, my memory usage will grow (more or less) without bound— the hit on memory is much worse than on compute. Or am I wrong about that?

I think in my original post, option 5 (something hacky like `Var::from_fake_global_symbol(usize)`) could definitely be used to avoid changing the user-facing API at all, and option 3 (simplifying `Var` so the global symbol `usize` values are reduced to/from `usize` where strings are input/output) could probably be done with no impact there at all (less sure about that).

Don't feel that I'm demanding or even expecting this change though— for the time being, I'm writing against my own fork of egg (a really bad habit of mine is hacking changes into upstream crates "for now"). So I'm ok, at least until the next round of improvements lands upstream. I just needed to figure out the "no strings attached" model for Egg for my own purposes, and I wanted to put the questions in the original issue to rest a bit.

mwillsey:
Coming back to this after a while: what about just allowing direct construction of Vars from `u32`, no interning or anything like that. Would that solve your use case? Would you need a way to get that number "back out" of the var?

robert-chiniquy:
Yep, that was more or less my option 5 from the discussion ticket. I would need to be able to get the usize back out. But that would do it. 

It's awesome that you're coming back to this. Hopeful for more egg in the future!

mwillsey:
I would love to keep the storage to `u32`, since that's what it's already
based on. Would that work for your use case?

robert-chiniquy:
Oh yeah absolutely, sorry just been writing usize a lot. Any non-FP number type is great.

Issue #113, opened 2021-08-05
Try de-duplicating parents


oflatt:
Right now, the parents list for each e-class doesn't de-duplicate congruent nodes. It's possible there would be a performance gain to doing so, since rebuilding would have to loop over fewer nodes the next iteration.

PR #291, opened 2024-01-03
Add nodes field to EGraph

dewert99:
I decided to try implementing the changes I described in #290 for adding a `nodes` field to `EGraph`. The main downside to my changes is that `EClass::parents` only returns non-canonical `Id`s and not the nodes themselves, although the nodes are available via `EGraph::id_to_node`. The main user-facing advantage is that `EGraph::id_to_X` methods as well as `EGraph::copy_without_unions` do not require explanations to be enabled. There are also memory usage/performance advantages since EClass.parents, EGraph.pending, and EGraph.analysis_pending, no longer need to store nodes. The nodes passed to `Analysis::make` are slightly different than before, but they are the same after canonicalization.

I also eliminated the `node` field of `ExplainNode` to reduce memory usage when explanations are enabled, but this did add some complexity. If this complexity isn't considered to be worth it, I could also revert the changes to `explain.rs` while preserving the rest of the changes.

### DIFF
--- a/src/egraph.rs
+++ b/src/egraph.rs
@@ -57,6 +57,8 @@ pub struct EGraph<L: Language, N: Analysis<L>> {
     /// The `Explain` used to explain equivalences in this `EGraph`.
     pub(crate) explain: Option<Explain<L>>,
     unionfind: UnionFind,
+    /// Stores the original node represented by each non-canonical id
+    nodes: Vec<L>,
     /// Stores each enode's `Id`, not the `Id` of the eclass.
     /// Enodes in the memo are canonicalized at each rebuild, but after rebuilding new
     /// unions can cause them to become out of date.
@@ -64,8 +66,8 @@ pub struct EGraph<L: Language, N: Analysis<L>> {
     memo: HashMap<L, Id>,
     /// Nodes which need to be processed for rebuilding. The `Id` is the `Id` of the enode,
     /// not the canonical id of the eclass.
-    pending: Vec<(L, Id)>,
-    analysis_pending: UniqueQueue<(L, Id)>,
+    pending: Vec<Id>,
+    analysis_pending: UniqueQueue<Id>,
     #[cfg_attr(
         feature = "serde-1",
         serde(bound(
@@ -114,6 +116,7 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
             analysis,
             classes: Default::default(),
             unionfind: Default::default(),
+            nodes: Default::default(),
             clean: false,
             explain: None,
             pending: Default::default(),
@@ -769,7 +772,9 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
                     *existing_explain
                 } else {
                     let new_id = self.unionfind.make_set();
-                    explain.add(original, new_id, new_id);
+                    explain.add(original.clone(), new_id, new_id);
+                    self.nodes.push(original);
+                    debug_assert_eq!(Id::from(self.nodes.len()), new_id);
                     self.unionfind.union(id, new_id);
                     explain.union(existing_id, new_id, Justification::Congruence, true);
                     new_id
@@ -778,7 +783,7 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
                 existing_id
             }
         } else {
-            let id = self.make_new_eclass(enode);
+            let id = self.make_new_eclass(enode, original.clone());
             if let Some(explain) = self.explain.as_mut() {
                 explain.add(original, id, id);
             }
@@ -791,24 +796,26 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
     }
 
     /// This function makes a new eclass in the egraph (but doesn't touch explanations)
-    fn make_new_eclass(&mut self, enode: L) -> Id {
+    fn make_new_eclass(&mut self, enode: L, original: L) -> Id {
         let id = self.unionfind.make_set();
         log::trace!("  ...adding to {}", id);
         let class = EClass {
             id,
             nodes: vec![enode.clone()],
-            data: N::make(self, &enode),
+            data: N::make(self, &original),
             parents: Default::default(),
         };
 
+        self.nodes.push(original);
+        debug_assert_eq!(Id::from(self.nodes.len()), id);
+
         // add this enode to the parent lists of its children
         enode.for_each(|child| {
-            let tup = (enode.clone(), id);
-            self[child].parents.push(tup);
+            self[child].parents.push(id);
         });
 
         // TODO is this needed?
-        self.pending.push((enode.clone(), id));
+        self.pending.push(id);
 
         self.classes.insert(id, class);
         assert!(self.memo.insert(enode, id).is_none());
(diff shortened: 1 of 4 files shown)


dewert99:
I noticed that since `EGraph::add_uncanonical` is optimized to avoid creating new nodes when an equivalent node is available, `EGraph::copy_without_unions` is incorrect when explanations aren't enabled, and the `EGraph::id_to_X` methods need to be clarified for the case where explanations are not enabled.

dewert99:
In order to get more stable benchmarks I tried using a [modified](https://github.com/iai-callgrind/iai-callgrind/pull/71) version of iai-callgrind:
[egraph_nodes-vs-main.txt](https://github.com/egraphs-good/egg/files/13865863/egraph_nodes-vs-main.txt)

I also compared `main` with itself to get a sense of stability:

[main-vs-main.txt](https://github.com/egraphs-good/egg/files/13865861/main-vs-main.txt)

mwillsey:
I like the idea here. @oflatt will know best about the interactions with proof stuff. Should we worry if the nodes Vec will grow too large (without proofs, in a long running example)? I guess we never compact the unionfind either, so at worst linear overhead I think.

dewert99:
Currently, this decreases memory usage even when not using explanations (at least in the benchmarks I ran) since the memory used by `nodes` is offset by not storing nodes in the parents list. This may change if we start compacting the parents lists (#113), but I think you're right about the linear overhead and I don't think the unionfind could be compacted. I guess the only problematic case is if the users' `Language` implementation is very large, they don't use explanations, and we do implement parent list compaction.

mwillsey:
This all kinds of begs the question: why not only store the e-nodes a single time? E-classes could refer to their nodes by id, and the memo could probably even get replaced with an `IndexMap`.

dewert99:
Replacing `memo` would be tricky since it uses canonicalized `nodes` and the EGraph `nodes` field uses un-canonicalized nodes. We could add a separate list of canonicalized nodes but I don't see how that would be any better keeping `memo` the way it is. The `nodes` field of `EClass` also gets canonicalized as part of re-building, although may not actually be necessary (just nice for deduplication). We would also need to experiment with the cost of the extra indirection when searching since currently the nodes field seems to be hit quite hard. Another idea I had that might help with this is to make `classes_by_op` into a `HashMap<L::Discriminant, HashMap<Id, u32>>` that indicates the first element of a classes `nodes` field that has a particular discriminant so we don't need to search from the beginning or binary search.

dewert99:
Another idea would be to combine `EGraph.nodes` and `Explain.uncanon_memo` into an `IndexSet<L>`, this would reduce the number of copies of nodes being use when explanations are enabled but add the overhead of the `RawTable` when explanations are not enabled

dewert99:
I was playing a bit with the idea of combining `EGraph.nodes` and `Explain.uncanon_memo` into an `IndexSet<L>` and I noticed that interestingly it is possible to have two ids that correspond to the exact same node (via `id_to_node`).

A somewhat minimal example of this is 
```rust
#[test]
fn dup_node() {
    use SymbolLang as S;

    crate::init_logger();
    let mut egraph = EGraph::<S, ()>::default().with_explanations_enabled();

    let y = egraph.add_uncanonical(S::leaf("y"));
    let y2 = egraph.add_uncanonical(S::leaf("y2"));
    egraph.add_uncanonical(S::new("f", vec![y]));
    egraph.add_uncanonical(S::new("g", vec![y2]));
    egraph.union(y, y2);
    egraph.add_uncanonical(S::new("f", vec![y]));
    egraph.add_uncanonical(S::new("g", vec![y2])); // adds the duplicate
    for i in 0..egraph.total_number_of_nodes() {
        let id = Id::from(i);
        println!("{id}: {:?} (root: {})", egraph.id_to_node(id), egraph.find(id))
    }
}
```
which prints:
```
0: SymbolLang { op: "y", children: [] } (root: 0)
1: SymbolLang { op: "y2", children: [] } (root: 0)
2: SymbolLang { op: "f", children: [0] } (root: 2)
3: SymbolLang { op: "g", children: [1] } (root: 3)
4: SymbolLang { op: "g", children: [1] } (root: 4)
```
This works before and after this PR (and after this PR it also works when explanations aren't enabled)

Basically what happens is that when calling `lookup_internal` inside of the second call to `egraph.add_uncanonical(S::new("g", vec![y2]))`, `y2` gets canonized to `y`, but since the entry in `memo` hasn't be updated yet it is still `S::new("g", vec![y2])` so `lookup_internal` returns `None` which forces a call to `make_new_eclass` which adds the duplicate node.

This isn't a bug since calling rebuild causes the duplicate nodes to get `union`ed.

This does give the idea of combining `EGraph.nodes` and `Explain.uncanon_memo` into an `IndexSet<L>`  an additional benefit that we can use it to double check that we don't add any duplicate nodes.

dewert99:
@mwillsey, I noticed you recently merged #306, so I updated PR to handle `EGraph` mapping, it actually simplifies it somewhat. Is this PR (as well as #296 and #300) still things you are interested in? I could also try to update #296 and #300 to handle egraph mapping if you would be interested in merging them.

src/explain.rs
```diff
@@ -1170,13 +1043,103 @@ impl<L: Language> Explain<L> {
         equalities
     }
 
-    pub(crate) fn populate_enodes<N: Analysis<L>>(&self, mut egraph: EGraph<L, N>) -> EGraph<L, N> {
-        for i in 0..self.explainfind.len() {
-            let node = &self.explainfind[i];
-            egraph.add(node.node.clone());
+    pub(crate) fn with_nodes<'a>(&'a mut self, nodes: &'a [L]) -> ExplainNodes<'a, L> {
```
oflatt, line 1046: `with_nodes` is a little awkward.
Another way to implement would be to keep the nodes in the `Explain` struct from the start. I think that might be cleaner?

src/egraph.rs
```diff
@@ -967,6 +1019,25 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
     /// assert_eq!(egraph.id_to_expr(fa), "(f a)".parse().unwrap());
     /// assert_eq!(egraph.id_to_expr(fb), "(f b)".parse().unwrap());
     /// ```
+    ///
```
oflatt, line 1022: Great documentation!

oflatt (approved):
Really great work! I'm in favor of this change. It cleans up the implementation nicely.
Did you ever get performance numbers? It's probably as performant as before.
Sorry I didn't review this sooner, and great job on this high-quality PR.

src/explain.rs
dewert99: The nodes need to be available when explanations are disabled so I couldn't only have them in the `Explain` struct like they were before, I didn't want to duplicate all of the nodes by having them in both places since that would require extra unnecessarily memory, and since rust doesn't support self-referential structs, I thought this was the easiest way to not change the explain implementation to much.

oflatt: Hmm, perhaps the Explain struct should be around whether explanations are disabled or enabled? Or I supposed the other fields of Explain could be wrapped in an `Option<OtherExplanationFields>`

dewert99: Do you mean have
```rust
struct Explain {
  nodes: Vec<L>,
  rest: Option<OtherExplanationFields>,
}
```
I don't think this would simplify things very much, and the nodes are use outside of explanations as well so I don't see why they should be part of the explain struct.
Some of the explain methods also take in the unionfind and classes, so I could also thread the nodes as an extra parameter if you prefer.

oflatt: Yeah, I see what you mean.
I don't have strong feelings, the solution you have is not too bad

dewert99: If you're fine with it, I think I'll just leave it as it is. Is there anything you would like me to change before I merge this?

oflatt: Nope, everything looks good to me

dewert99: Great, Thanks

src/egraph.rs
```diff
@@ -1324,14 +1400,15 @@ impl<L: Language, N: Analysis<L>> EGraph<L, N> {
                 }
             }
 
-            while let Some((node, class_id)) = self.analysis_pending.pop() {
+            while let Some(class_id) = self.analysis_pending.pop() {
+                let node = self.nodes[usize::from(class_id)].clone();
                 let class_id = self.find_mut(class_id);
                 let node_data = N::make(self, &node);
```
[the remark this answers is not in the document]
dewert99: > Does this mean that make can be called with un-canonical e-class ids?

Yes, but I think this was also true before this PR, I don't think we ever canonized the nodes in `analysis_pending`, or `parents`.

> If so we should at least say so in the make docs.

If you want I can still add that

mwillsey:
Yes, please. Then I'm good to merge!

PR #299, opened 2024-02-11
[doc] redundant web demo

majidaldo:


mwillsey:
Why is the web demo redundant?

majidaldo:
> Why is the web demo redundant?

there are two links to a web demo. one demo seems to be an old version...the one i deleted.

mwillsey:
Those are two different demos. One is of egg (the one this PR deletes), the other is in a paragraph describing the newer egglog tool, and a demo for that. Despite egg being older, I think the demo still has some value in explaining how it works.

majidaldo:
> Those are two different demos. One is of egg (the one this PR deletes), the other is in a paragraph describing the newer egglog tool, and a demo for that. Despite egg being older, I think the demo still has some value in explaining how it works.

...so it should be explained in the readme.

mwillsey:
Done!

Issue #305, opened 2024-03-20
Map from `EGraph<L, _>` to `EGraph<L2, _>`


sgpthomas:
I want to use the [babble](https://github.com/dcao/babble) library with an e-graph that I own. `babble` expects an EGraph over the `AstNode<Op>` (which implements `egg::Language`) type. I have an EGraph over my own language type, let's call it `Lang`, and it's trivial to write a function from `Lang -> AstNode<Op>`. However, as far as I can tell, there is not a good way to lift this into a function `EGraph<Lang, _> -> `EGraph<AstNode<Op>, _>`. Am I missing something in the library, or do you have a recommendation for how to do this?

mwillsey:
Closed by #306

PR #306, opened 2024-03-20
Translate between EGraph types

sgpthomas:
Addresses issue #305.

I use some pretty complicated trait bounds in the `EGraph::map` function to handle converting between analyses. If the analysis stays the same (`EGraph<L, T> -> EGraph<L2, T>`) these are automatically implemented which makes it more convenient to use in what I imagine is the common case. However, let me know if you think the conversion should be explicitly handled with a passed in closure.

### DIFF
--- a/src/eclass.rs
+++ b/src/eclass.rs
@@ -41,6 +41,26 @@ impl<L, D> EClass<L, D> {
     pub fn parents(&self) -> impl ExactSizeIterator<Item = (&L, Id)> {
         self.parents.iter().map(|(node, id)| (node, *id))
     }
+
+    /// Converts an EClass defined for language L into one
+    /// defined for language L2
+    pub fn map<F, L2, D2>(self, mapper: F) -> EClass<L2, D2>
+    where
+        F: Fn(L) -> L2,
+        D2: From<D>,
+        L2: Language,
+    {
+        EClass {
+            id: self.id,
+            nodes: self.nodes.into_iter().map(&mapper).collect(),
+            data: self.data.into(),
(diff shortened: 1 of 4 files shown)


mwillsey:
Here's an alternative idea that might ease some of the trait bounds, and it would make things a little more flexbile than requiring `From`. Make `LangMapper` a trait or something:

```rust
trait LangMapper<L, A> {
  type L2: Language;
  type A2: Analysis<Self::L2>;
  fn map_node(&self, node: L) -> Self::L2;
  fn map_analysis(&self, ...)  {
    // should maybe be able to provide a default impl?
  }
  fn map_egraph(&self, egraph: EGraph<L, A>) -> EGraph<L2, A2>
}
```

If you wanted to go crazy, those methods could all take in the src and dst egraphs? What do you think?

sgpthomas:
I liked that idea and implemented it. I couldn't figure out a nice way to provide default implementations in the `LanguageMapper` trait, so I wrote a `SimpleLanguageMapper` struct that provides a `LanguageMapper` implementation when there are the necessary `From<..>` impls. Those type constraints are still pretty gnarly, but they aren't on the default interface. Let me know what you think of that.

mwillsey:
Looks good to me! Thanks for the contribution. Seems like some doctests are failing. Could you patch that up? Try `cargo test --doc`. Btw `make test` does everything the CI does.

sgpthomas:
Okay, should be good to go

mwillsey:
Thanks for the contribution!

PR #308, opened 2024-04-02
Add example in doc comment of 'RecExpr'

marcusrossel:
As requested by @oflatt in https://egraphs.zulipchat.com/#narrow/stream/375765-egg.2Fegglog/topic/Replace.20Var.20in.20Pattern/near/430172065

### DIFF
--- a/src/language.rs
+++ b/src/language.rs
@@ -369,7 +369,9 @@ impl LanguageChildren for Id {
 /// a list of enodes.
 ///
 /// [`RecExpr`]s must satisfy the invariant that enodes' children must refer to
-/// elements that come before it in the list.
+/// elements that come before it in the list. For example, the expression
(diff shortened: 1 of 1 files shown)


oflatt (approved):
Thanks!

oflatt (changes requested):
Looks like you fail the CI (I had to approve the run, I'll change the settings so it always runs in the future)

marcusrossel:
Sorry, I don't understand what the problem is right now. Is there anything I need to do?

saulshanabrook:
I think the formatter is complaining about some whitespace: https://github.com/egraphs-good/egg/actions/runs/8525410185/job/23363979154#step:4:1

If you run "cargo fmt" locally on your branch it should fix it.

PR #312, opened 2024-04-24
Return `Language::len` in O(1)

cospectrum:
`self.children().len()` instead of `self.children().iter().fold(0, |len, _| len + 1)`


cospectrum:
This may have been done intentionally from the start. For example, if user implements fold recursively. It's better to leave everything as is, and users who need speed implement it themselves.

PR #313, opened 2024-04-24
Add `hash_map_with_capacity`

cospectrum:
I'm using a separate function, because `HashMap::with_capacity(cap)` doesn't work for `hashbrown` and `fxhash`.

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -1,7 +1,7 @@
 use std::cmp::Ordering;
 use std::fmt::Debug;
 
-use crate::util::HashMap;
+use crate::util::{hash_map_with_capacity, HashMap};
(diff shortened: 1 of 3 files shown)


mwillsey:
Why not just say `HashMap::with_capacity_and_hasher(cap, util::BuildHasher::default())`?

cospectrum:
> Why not just say `HashMap::with_capacity_and_hasher(cap, util::BuildHasher::default())`?

Too long, takes 2 lines, it's a matter of taste

mwillsey:
Fair enough. Please change to `hashmap_with_capacity` for consistency and then we can merge.

cospectrum:
> Fair enough. Please change to `hashmap_with_capacity` for consistency and then we can merge.

Done

Issue #37, opened 2020-07-16
Rename RecExpr to Expr


mwillsey:
It's not recursive anymore.

PR #314, opened 2024-04-24
Rename `RecExpr` to `Expr`

cospectrum:
Renamed everywhere except the changelog.
`recexpr` -> `expr`.
`RecExpr` -> `Expr`.
Increased the crate version to `0.10.0` in Cargo.toml.

To close https://github.com/egraphs-good/egg/issues/37


mwillsey:
Thanks for the contribution, but I think we'll keep RecExpr for now to avoid breakage.

PR #320, opened 2024-06-24
chore: keep rust version up-to-date with `stable` channel in CI

kayagokalp:
# Description

Keeps rust version up-to-date with `stable` channel in CI. It is also possible to pin to a version. Since this makes the `rustup` follow stable channel all the time it should not break anything but nits might appear so some maintenance might be required to keep up-to-date from time to time.  Also moves runner to `ubuntu-latest`, `docs.yml` is already using `ubuntu-latest`.


kayagokalp:
unblocks #318

kayagokalp:
superseded by [dae3050](https://github.com/egraphs-good/egg/commit/dae30504dd45c671cc8840b76c4265db99d3aa54)

PR #321, opened 2024-07-04
Added `fn root(&self) -> Id` to `RecExpr`.

yonatan-reicher:
This is a convenience method for getting the root node of an expression for working on `RecExpr` directly without needing to know that it is the last element of it's slice.


asukaminato0721:
> `error: package `syn v2.0.68` cannot be built because it requires rustc 1.61 or newer, while the currently active rustc version is 1.60.0`

PR #324, opened 2024-07-29
remove hashbrown from id_to_pattern API

mtak-:
Hello! I've been having so much fun with `egg`, I thought I would contribute a little.

I've replaced a `HashMap` with a `FnMut` in the API. I tried using the `HashMap` version and found I had to use the same verison of hashbrown with fxhash as egg.

This is a breaking API change and requires a version bump.


mwillsey:
Not a bad idea, but I'm okay with requiring the use of hashbrown and the correct hashfn. Thanks though!

PR #325, opened 2024-07-29
RecExpr Deref/DerefMut=[L]. also convenience methods and trait impls

mtak-:
I believe the current API of RecExpr already constrains the impl enough to go ahead and add `Deref/DerefMut = [L]`. I could instead add `len`/`iter`/`iter_mut` and still get most of the benefits of `Deref/DerefMut`.

I also added `FromIterator`, `AsMut`, `Borrow`, `BorrowMut`, and `IntoIterator` impls, and a couple of convenience methods. `items` solves the issue of `expr.as_ref().iter().enumerate()` followed by `Id::from(idx)`

I went ahead and changed the source to take advantage of these additions. Let me know if you would like me to scale this pr back.


mwillsey:
Nice change!

PR #330, opened 2024-08-09
Parallel matching

szymon-zygula:
This PR adds a parallel matching algorithm using relaxed memory ordering atomic operations in the `UnionFind`, allowing for parallel path compression. Theoretically, this is overhead-free on x86 (because of total store order), other than some cache coherence magic. Some people may remember it from conversations at PLDI 2024. This makes parallel matching scale well with the number of threads.

The functionality is hidden behind `parallel-matching` feature. When the feature is activated, you can set `parallel_matching` to `true` in `Runner` to use it.

Atomics in the `UnionFind` are always used, regardless of the `parallel-matching` feature. There could be two implementations, the new one and the old one, depending on `parallel-matching`, although in my opinion it would add unnecessary maintenance overhead. Opinions on this topic are welcome.

### DIFF
--- a/src/language.rs
+++ b/src/language.rs
@@ -28,11 +28,11 @@ use thiserror::Error;
 ///
 /// See [`SymbolLang`] for quick-and-dirty use cases.
 #[allow(clippy::len_without_is_empty)]
-pub trait Language: Debug + Clone + Eq + Ord + Hash {
+pub trait Language: Debug + Clone + Eq + Ord + Hash + Send + Sync {
     /// Type representing the cases of this language.
     ///
     /// Used for short-circuiting the search for equivalent nodes.
-    type Discriminant: Debug + Clone + Eq + Hash;
+    type Discriminant: Debug + Clone + Eq + Hash + Send + Sync;
 
     /// Return the `Discriminant` of this node.
     #[allow(enum_intrinsics_non_enums)]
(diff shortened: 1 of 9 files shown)


src/language.rs
```diff
@@ -28,11 +28,11 @@ use thiserror::Error;
 ///
 /// See [`SymbolLang`] for quick-and-dirty use cases.
 #[allow(clippy::len_without_is_empty)]
-pub trait Language: Debug + Clone + Eq + Ord + Hash {
+pub trait Language: Debug + Clone + Eq + Ord + Hash + Send + Sync {
```
mwillsey, line 31: I don't think it's acceptable to change the bounds of `Language` at this point in `egg`'s lifecycle. Getting rid of this I think is the main challenge the merging this.

mwillsey (changes requested):
Hi, thanks for the contribution!

I like the idea of merging this, but I'm not willing to accept any breaking changes for it (when you aren't using the feature). I also think that the unionfind change isn't necessary to parallel matching; you should just be able to us the existing one.

Thoughts?

src/language.rs
szymon-zygula: I added a new trait `MaybePar`, which is
- automatically implemented for every type when `parallel-matching` is not set,
- automatically implemented for `Sync + Send` types when `parallel-matching` is not set.
`Language`, `Analysis`, `Analysis::Data`, and some other traits now have bounds involving this trait, so nothing changes when `parallel-matching` isn't set. Let me know what you think about this solution. If [trait aliases](https://doc.rust-lang.org/unstable-book/language-features/trait-alias.html) ever get stabilized, they could be used instead.

mwillsey:
I think this is a cool contribution, but I will close it for now in favor of this commit I just added: https://github.com/egraphs-good/egg/commit/d0148001e021a4a1faf9a21ceed8271f4e5c789c.

It should be quite easy to add (rule-based) parallelism on the client side without any further changes to egg. I have documented an example of doing so.

PR #333, opened 2024-08-20
Remove Existence Explanations

oflatt:
Removed existence explanations from egg (the `explain_existance` function). This feature was buggy and not well supported. Supporting it fully required many changes, and it is incompatible with analysis. See #332 for more details.


oflatt:
@eytans  is this alright with you?

PR #337, opened 2024-10-22
Add data field in runner to solve hook ownership issues

Bastacyclop:
I ran into the following problem a couple of times: you try to use some owned value inside the `Runner` hook, but you can't afford to move it into the `'static` lifetime closure.

Here's a simple  example:
```rust
fn runner_issue<W>(mut out: W)
    where W: std::io::Write,
{
    let runner = Runner::<SymbolLang, ()>::default()
        .with_hook(move |r| {
            writeln!(out, "{} iterations", r.iterations.len()).unwrap();
            Ok(())
        });

    writeln!(out, "{} iterations", runner.iterations.len()).unwrap();
}
```

Which gives the following error:
```
> cargo build
   Compiling egg v0.9.5 (/home/thomas/egg)
error[E0310]: the parameter type `W` may not live long enough
    --> src/run.rs:1009:18
     |
1009 |       let runner = Runner::<SymbolLang, ()>::default()
     |  __________________^
1010 | |         .with_hook(move |r| {
1011 | |             writeln!(out, "{} iterations", r.iterations.len()).unwrap();
1012 | |             Ok(())
1013 | |         });
     | |          ^
     | |          |
     | |__________the parameter type `W` must be valid for the static lifetime...
     |            ...so that the type `W` will meet its required lifetime bounds
     |
help: consider adding an explicit lifetime bound
     |
1007 |     where W: std::io::Write + 'static,
     |                             +++++++++

error[E0382]: borrow of moved value: `out`
    --> src/run.rs:1015:14
     |
1006 | fn runner_issue<W>(mut out: W)
     |                    ------- move occurs because `out` has type `W`, which does not implement the `Copy` trait
...
1010 |         .with_hook(move |r| {
     |                    -------- value moved into closure here
1011 |             writeln!(out, "{} iterations", r.iterations.len()).unwrap();
     |                      --- variable moved due to use in closure
...
1015 |     writeln!(out, "{} iterations", runner.iterations.len()).unwrap();
     |              ^^^ value borrowed here after move
     |
help: if `W` implemented `Clone`, you could clone the value
    --> src/run.rs:1006:17
     |
1006 | fn runner_issue<W>(mut out: W)
     |                 ^ consider constraining this type parameter with `Clone`
...
1011 |             writeln!(out, "{} iterations", r.iterations.len()).unwrap();
     |                      --- you could clone this value
help: consider further restricting this bound
     |
1007 |     where W: std::io::Write + Copy,
     |                             ++++++
```

This PR adds a custom data field to the `Runner`, so that any value can be moved in and out of the runner, and therefore in and out of the hook closure:

```rust
fn runner_issue_solved<W>(out: W)
    where W: std::io::Write,
{
    let mut runner = Runner::<SymbolLang, (), (), W>::new((), out)
        .with_hook(move |r| {
            writeln!(r.data, "{} iterations", r.iterations.len()).unwrap();
            Ok(())
        });

    writeln!(runner.data, "{} iterations", runner.iterations.len()).unwrap();
}
```

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -99,9 +99,9 @@ pub struct MyIterData {
     smallest_so_far: usize,
 }
 
-type MyRunner = Runner<SimpleLanguage, (), MyIterData>;
+type MyRunner = Runner<SimpleLanguage, (), MyIterData, ()>;
 
-impl IterationData<SimpleLanguage, ()> for MyIterData {
+impl IterationData<SimpleLanguage, (), ()> for MyIterData {
     fn make(runner: &MyRunner) -> Self {
         let root = runner.roots[0];
         let mut extractor = Extractor::new(&runner.egraph, AstSize);
@@ -134,9 +134,11 @@ println!(
 
 ```
 */
-pub struct Runner<L: Language, N: Analysis<L>, IterData = ()> {
+pub struct Runner<L: Language, N: Analysis<L>, IterData = (), Data = ()> {
     /// The [`EGraph`] used.
     pub egraph: EGraph<L, N>,
+    /// Data stored inside the runner, useful for making that data available to hooks
+    pub data: Data,
     /// Data accumulated over each [`Iteration`].
     pub iterations: Vec<Iteration<IterData>>,
     /// The roots of expressions added by the
@@ -189,7 +191,7 @@ impl RunnerLimits {
     }
 }
 
-impl<L, N> Default for Runner<L, N, ()>
+impl<L, N> Default for Runner<L, N, (), ()>
 where
     L: Language,
     N: Analysis<L> + Default,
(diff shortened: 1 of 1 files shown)


Bastacyclop:
To the best of my knowledge there is no satisfying way of going from `'a` lifetime to `'static` lifetime, this corresponds to a memory leak, e.g. https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak
None of Rc, RefCell, Arc or Mutex can do this: what they allow is sharing with interior mutability.
Could you write a fix that compiles for the code above without adding an overly restrictive `+ 'static` constraint to `W` ?

Bastacyclop:
The alternative fix would be to play with lifetime parameters in Runner instead of type parameters, so that the hook closure is not `'static`. That seems more complicated than adding a type parameter to me.

Bastacyclop:
For posteriority, the suggested workaround is to change the `runner_issue` signature to force callers to pass in only 'static + Clone writers which they may achieve, on their side, through Rc etc:
```rust
fn runner_issue_workaround<W>(mut out: W)
    where W: std::io::Write + 'static + Clone,
{
    let mut out_captured = out.clone();
    let runner = Runner::<SymbolLang, ()>::default()
        .with_hook(move |r| {
            writeln!(out_captured, "{} iterations", r.iterations.len()).unwrap();
            Ok(())
        });

    writeln!(out, "{} iterations", runner.iterations.len()).unwrap();
}
```
So that workaround is **not** enabling to transparently go from 'a to 'static, instead it is about ensuring 'static lifetime from the get go.

PR #341, opened 2024-12-25
Use $crate::Id to reference the egg::Id type in `define_language!`

jungnitz:
In the `define_language!` macro, there is a small oversight where `Id` is used without the `$crate::`-prefix


mwillsey:
Thanks!

PR #346, opened 2025-01-21
EGraph minimization with bisimulation

danieltherealyang:
Implemented bisimulation to convert egraph to dtfa, minimize the dtfa, then convert the dtfa back to egraph. There is quite a bit of overhead but the code is all self contained. If someone wants to try optimizing the minimization code they can do that too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ vectorize = { version = "0.2.0", optional = true }
 # for the reports feature
 serde_json = { version = "1.0.81", optional = true }
 saturating = "0.1.0"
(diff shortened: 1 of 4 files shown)


oflatt:
Very cool
I don't know if you've seen this repo: https://github.com/egraphs-good/egraph-serialize
It has a bunch of egraphs you could try your algorithm on, and see if you can improve the extraction results or performance

mwillsey:
This is very cool! But I don't currently think it should be merged. It seems all of this could be implemented outside off egg in a separate crate, which is my preference.

Issue #350, opened 2025-03-03
Conditions in explanations


adrianleh:
Hi,

I'm currently using explanations on a set of rewrites, of which some are conditional. However, when I get the proof explanation all rewrite steps to rewrite my terms are listed, except the ones that fulfilled the conditions of conditional rewrites. Is it possible to force a condition to include the explanation of the existence of the matched enode into the explanation?

oflatt:
Unfortunately, egg's explanations system does not support conditional rewrites. We are actively trying to add such support to egglog, but it's turning out to be a tough problem.

If it's possible to change the encoding to include the condition as part of the term, that might be a work around with some effort. You could also include in the term some hints that let whatever you're using with the result reconstruct the proof. Hope this helps.

adrianleh:
Thanks for the feedback. Could you elaborate a bit on the problem here. In my understanding it seems to me that typically conditions prove the existence of an enode or the equality of two enodes, which the checker could handle? Also is there an updated date on the `egglog` proof engine?

oflatt:
Egg's current proof generation algorithm and format doesn't support multiple preconditions- it was built with the assumption of matching a single term.

No solid date on the egglog proof engine, but we are focusing more on it now.

adrianleh:
I noticed that but it's quite easy to hack in multiple preconditions by creating custom AndConditions that take multiple conditions and ensure they are all met

oflatt:
Sure, open a PR and I'll take a look. Happy to take a look at your code, if we change egg we'd want to consider carefully. I'd like to see how it interacts with the various proof formats and what kinds of conditions are supported.

PR #351, opened 2025-03-07
Add multi-condition rewrites

adrianleh:
As discussed in #350, adding simple wrapper structures to allow for rewrites to be conditioned on multiple conditions

(cc: @oflatt)

### DIFF
--- a/src/rewrite.rs
+++ b/src/rewrite.rs
@@ -538,6 +538,105 @@ where
     }
 }
 
+#[derive(Clone)]
+pub struct AndCondition<L, N>
+where
+    L: Language,
+    N: Analysis<L>,
+{
+    pub c1: Arc<dyn Condition<L, N> + Send + Sync>,
+    pub c2: Arc<dyn Condition<L, N> + Send + Sync>,
+}
+
+impl<L: Language, N: Analysis<L>> Condition<L, N> for AndCondition<L, N> {
+    #[inline]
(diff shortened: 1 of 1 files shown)


src/rewrite.rs
```diff
@@ -538,6 +538,105 @@ where
...
+    conds
+        .into_iter()
+        .fold(Arc::new(FalseCondition {}), |acc, c| {
+            Arc::new(OrCondition { c1: acc, c2: c })
+        })
+}
+
+pub fn all<L, N>(
+    conds: Vec<Arc<dyn Condition<L, N> + Send + Sync>>,
+) -> Arc<dyn Condition<L, N> + Send + Sync>
```
adrianleh, line 628: We could also easily change the return type here into `Arc<AndCondition<L,N>>` if we wanted to (same for `any` with `OrCondition`) by making the base an and of true conditions. Little less efficient but nicer types

oflatt:
How does this interact with proofs? I thought proofs didn't support conditions at all.
Do you have an example proof?

oflatt:
Oh I see, I think I misunderstood in the issue thread.
These seem useful enough to merge. What do you think @mwillsey?
We'll want documentation and some small tests before merging.

mwillsey:
Sorry for the delay here. Maybe I'm misunderstanding, but can't you already achieve multiple conditions via stacking conditional appliers?

PR #352, opened 2025-03-13
Change analysis make function to also take the Id of the created class.

pajampauls:
I have added the Id of the newly created class on calls of analysis make. This allows users to also process the Id in make, where previously the node would not yet return its class when calling egraph.lookup(node) during make.


mwillsey:
Nice change. I did a very similar one with a `remake` function, so you can tell the difference between initial make and rebuilds if that's useful.

PR #356, opened 2025-08-28
Migrate LpExtract to `good_lp` to support many ILP solvers.

recmo:
Conveniently, the default solver in `good_lp` is `coin_cbc`, so nothing changes unless a different backend is chosen.

Two interesting solvers this enables are

- [**HiGHS**](https://docs.rs/microlp/latest/microlp/) which has excellent performance and does not require installing anything.
- [**microlp**](https://docs.rs/microlp/latest/microlp/) which is rust-native and can compile to webassembly.

This PR also includes the commits from #355, so you may want to merge that first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ symbolic_expressions = "5.0.3"
 thiserror = "1.0.31"
 
 # for the lp feature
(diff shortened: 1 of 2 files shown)


mwillsey:
Thanks!

PR #357, opened 2025-09-01
Remove unused saturating dependency

Tpt:
Thank you so much for this great library!

The `saturating` seems to be unused


mwillsey:
Thanks!

PR #364, opened 2026-04-08
feat: add no_std + alloc support behind std feature gate

### LINKED ISSUE #361: Suggestion: `no_std` support (with `alloc`)
I am writing a compiler for WASM and want to use e-graphs (particularly this crate) to optimize the output. However, the compiler is `no_std`, so this crate would need to be ported to be `no_std` compatible to work in it. Using `alloc` would be a possibility to start, as we don't use any OS functions in `egg` afaik.

MavenRain:
Port the entire crate to work without the standard library by gating std-only dependencies (env_logger, quanta, symbol_table, symbolic_expressions) behind a default "std" feature.  When std is disabled, a minimal sexp parser, a spin-lock-based global symbol interner, and a no-op Instant shim fill in.  Bumps thiserror to v2 for no_std error derive support.

Closes #361

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,18 +11,19 @@ repository = "https://github.com/egraphs-good/egg"
 version = "0.11.0"
 
 [dependencies]
-env_logger = {version = "0.9.0", default-features = false}
-hashbrown = "0.15.2"
-indexmap = "2.7.0"
-log = "0.4.17"
-num-bigint = "0.4"
-num-traits = "0.2"
-quanta = "0.12"
-rustc-hash = "2.0.0"
+env_logger = {version = "0.9.0", default-features = false, optional = true}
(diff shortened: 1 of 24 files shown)


mwillsey:
Hi, this is an interesting contribution, but it's not ready. To merge this, I'd like to consolidate most import changes into a shim thing in `lib.rs`. `#[allow(unused_imports)]` should not be used. Also, this patch doesn't actually test any of the changes. Either manually run `cargo test --no-default-features`, or add it to the Makefile.

mwillsey:
Looks good. I've added no_std support to the symbol_table library. Please depend on it's latest commit and then looks should be good to go. https://github.com/mwillsey/symbol_table

mwillsey:
Thanks!

